PR #33, opened 2026-04-08
refactor: simplify complex codepaths

### LINKED ISSUE #32: refactor: simplify complex codepaths
## Summary

Audit of complex codepaths across server and agent identified three high-value simplification opportunities and several that aren't worth changing.

---

### 1. Duplicate serialization functions in action_handler.go (Server)

`serializeCreateActionParams` (lines 1153-1208) and `serializeUpdateActionParams` (lines 1210-1265) are **identical 55-line functions** — both marshal a proto oneof field to JSON via `protojson.Marshal`, then unmarshal to `map[string]any`. The only difference is the source type (`CreateActionRequest` vs `UpdateActionParamsRequest`).

**Fix:** Replace both with a single generic function:
```go
func serializeProtoParams(msg proto.Message) (map[string]any, error) {
    if msg == nil {
        return map[string]any{}, nil
    }
    data, err := protojson.Marshal(msg)
    if err != nil {
        return nil, fmt.Errorf("marshal params: %w", err)
    }
    var params map[string]any
    if err := json.Unmarshal(data, &params); err != nil {
        return nil, fmt.Errorf("unmarshal params to map: %w", err)
    }
    return params, nil
}
```

Callers extract the concrete proto message from the oneof and pass it directly. The 18-case switch statements in both functions collapse to a single call. Also eliminates `serializeActionParamsToMap` (lines 1273-1328) which is a third copy with silent error swallowing.

**Impact:** ~170 lines → ~20 lines

### 2. handleAgentMessage in handler/agent.go (Server)

251-line function with 11 cases, each following the same pattern: extract payload → build task payload → enqueue to control. The ActionResult case alone is 57 lines with nested JSON parsing and LPS rotation handling.

**Fix:** Split each case into a named method:
```go
func (h *AgentHandler) handleAgentMessage(ctx context.Context, deviceID string, msg *pm.AgentMessage) error {
    switch p := msg.Payload.(type) {
    case *pm.AgentMessage_Heartbeat:
        return h.handleHeartbeat(ctx, deviceID, p.Heartbeat)
    case *pm.AgentMessage_ActionResult:
        return h.handleActionResult(ctx, deviceID, p.ActionResult)
    // ...
    }
}
```

Each handler method becomes 10-30 lines with a single responsibility. The ActionResult handler can further extract LPS rotation proxying into its own method.

**Impact:** One 251-line function → eleven 10-30 line methods. No logic change, just structure.

### 3. executePackage in executor/executor.go (Agent)

167-line function with 7-level nesting in the PRESENT state path. The version check → pin check → install flow creates a deeply nested if-else tree that's hard to follow.

**Fix:** Flatten with guard clauses and extract sub-functions:
```go
func (e *Executor) executePackage(...) {
    switch state {
    case PRESENT:
        return e.ensurePackagePresent(ctx, params, pkgName)
    case ABSENT:
        return e.ensurePackageAbsent(ctx, params, pkgName)
    }
}

func (e *Executor) ensurePackagePresent(...) {
    isInstalled, _ := e.pkgManager.IsInstalled(pkgName)
    if !isInstalled {
        return e.installPackage(ctx, params, pkgName)
    }
    // Already installed — check version and pin
    if params.Version != "" {
        return e.ensurePackageVersion(ctx, params, pkgName)
    }
    if params.Pin {
        return e.ensurePackagePinned(ctx, pkgName)
    }
    return alreadyInstalled(pkgName), false, nil
}
```

**Impact:** One 167-line function with 7-level nesting → four 20-40 line functions with max 2-level nesting.

---

## Not worth changing

These were evaluated and intentionally skipped:

| Function | Why it looks complex | Why it's fine |
|----------|---------------------|---------------|
| `validateCreate/UpdateActionParams` | 18-case switch | Inherent proto oneof dispatch — Go type switches are the idiomatic pattern. Can't table-drive without reflection, which would be worse. |
| `ExecuteWithStreaming` (agent) | 19-case switch dispatch | Straightforward routing. Each case is 2-5 lines. The switch IS the dispatch table. |
| `main()` (server + agent) | 450+ / 180+ lines | Init sequences are inherently sequential. Extracting to functions adds indirection without reducing cognitive load — you'd just be jumping between files instead of reading top-to-bottom. |
| `DispatchAction` | 158 lines | Mixed concerns (validation, DB, queue) but each section is clearly commented and sequential. Splitting would scatter related logic. |
| `handleExecutionResult` | 195 lines | Already simplified in #30. Remaining complexity is inherent (ID resolution + status mapping + compliance check). |

PaulDotterer:
## Summary

Simplifies the two most complex server codepaths identified in #32.

### 1. Consolidate serialization functions (-29 lines)

Replaced three near-identical functions with a single `serializeProtoParams(proto.Message)`:
- `serializeCreateActionParams` (55 lines, 18-case switch)
- `serializeUpdateActionParams` (55 lines, 18-case switch)
- `serializeActionParamsToMap` (55 lines, 16-case switch with silent error swallowing)

Callers now extract the concrete `proto.Message` from the oneof via type-specific extractors (`extractCreateActionParamsMsg`, `extractUpdateActionParamsMsg`, `extractActionParamsMsg`) and pass it to the shared serializer.

### 2. Split handleAgentMessage (265 lines → 11 methods)

The monolithic switch with 11 cases is now a dispatcher routing to focused methods:
- `handleHeartbeat` (12 lines)
- `handleActionResult` (25 lines) + `proxyLpsRotations` (30 lines)
- `handleOutputChunk` (18 lines)
- `handleQueryResult` (18 lines)
- `handleInventory` (20 lines)
- `handleSecurityAlert` (12 lines)
- `handleGetLuksKey` (16 lines)
- `handleStoreLuksKey` (18 lines)
- `handleRevokeLuksResult` (12 lines)
- `handleLogQueryResult` (12 lines)

No behavior change — same logic, same error handling, same return values.

### Item 3 (agent executePackage)

Already implemented in manchtools/power-manage-agent#34.

Closes #32

## Test plan
- [x] `go build -o /dev/null ./cmd/control && go build -o /dev/null ./cmd/gateway`
- [x] `go vet ./...`
- [x] `go test ./internal/handler/... ./internal/gateway/...`
- [ ] Full `go test ./...` in CI (requires PostgreSQL)

## Summary by CodeRabbit

* **Refactor**
  * Streamlined parameter serialization to be more consistent and reliable across actions.
  * Reorganized agent message processing into a clearer dispatch flow, improving maintainability and reducing complexity.

### DIFF
--- a/internal/api/action_handler.go
+++ b/internal/api/action_handler.go
@@ -661,7 +661,10 @@ func (h *ActionHandler) DispatchAction(ctx context.Context, req *connect.Request
 		action := source.InlineAction
 		actionType = action.Type
 		desiredState = action.DesiredState
-		params, _ = serializeProtoParams(extractActionParamsMsg(action))
+		params, err = serializeProtoParams(extractActionParamsMsg(action))
+		if err != nil {
+			return nil, apiErrorCtx(ctx, ErrValidationFailed, connect.CodeInvalidArgument, fmt.Sprintf("failed to serialize inline action params: %v", err))
+		}
 		timeoutSeconds = action.TimeoutSeconds
 		if timeoutSeconds <= 0 {
 			timeoutSeconds = 300
@@ -1289,6 +1292,8 @@ func extractActionParamsMsg(action *pm.Action) proto.Message {
 		return p.Lps
 	case *pm.Action_Luks:
 		return p.Luks
+	case *pm.Action_Group:
+		return p.Group
 	case *pm.Action_Wifi:
 		return p.Wifi
 	case *pm.Action_AgentUpdate:
--- a/internal/api/action_handler.go
+++ b/internal/api/action_handler.go
@@ -13,6 +13,7 @@ import (
 	"github.com/jackc/pgx/v5"
 	"github.com/oklog/ulid/v2"
 	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	pm "github.com/manchtools/power-manage/sdk/gen/go/pm/v1"
@@ -257,7 +258,7 @@ func (h *ActionHandler) CreateAction(ctx context.Context, req *connect.Request[p
 		return nil, err
 	}
 
-	params, err := h.serializeCreateActionParams(req.Msg)
+	params, err := serializeProtoParams(extractCreateActionParamsMsg(req.Msg))
 	if err != nil {
 		return nil, apiErrorCtx(ctx, ErrValidationFailed, connect.CodeInvalidArgument, err.Error())
 	}
@@ -484,7 +485,7 @@ func (h *ActionHandler) UpdateActionParams(ctx context.Context, req *connect.Req
 		return nil, handleGetError(ctx, err, ErrActionNotFound, "action not found")
 	}
 
-	params, err := h.serializeUpdateActionParams(req.Msg)
+	params, err := serializeProtoParams(extractUpdateActionParamsMsg(req.Msg))
 	if err != nil {
 		return nil, apiErrorCtx(ctx, ErrValidationFailed, connect.CodeInvalidArgument, err.Error())
 	}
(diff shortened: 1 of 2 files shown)


PaulDotterer:
@coderabbitai full review

PaulDotterer:
@coderabbitai review

PR #34, opened 2026-04-11
feat: stub remote terminal RPCs on ControlService

PaulDotterer:
## Summary

Coordinated companion to manchtools/power-manage-sdk#25 (SDK proto contract for the remote terminal feature). Without this PR, the server stops compiling once the SDK update lands because \`*ControlService\` no longer satisfies \`pmv1connect.ControlServiceHandler\`.

This PR adds no-op stub implementations for the 4 new RPCs that return \`CodeUnimplemented\` via the existing \`apiErrorCtx\` helper. Real implementations land as part of step 5 of manchtools/power-manage-sdk#16: TTY user provisioning, session token minting, gateway URL resolution, and admin session management.

## What changes

\`internal/api/service.go\` — 4 new methods at the bottom of the file, following the existing one-method-per-RPC pattern:

- \`StartTerminal\`
- \`StopTerminal\`
- \`ListActiveTerminalSessions\`
- \`TerminateTerminalSession\`

All four return:

\`\`\`go
return nil, apiErrorCtx(ctx, ErrUnimplemented, connect.CodeUnimplemented, "remote terminal sessions are not yet implemented")
\`\`\`

so the web client gets a clear, structured error code rather than a panic or "method not found" from Connect-RPC.

## Why now

Issue manchtools/power-manage-server#6 (Remote Shell / Live Terminal) is the server-side counterpart of manchtools/power-manage-sdk#16. The proto contract is shared between them and was deliberately split out from both feature implementations because it has to land first — every other piece (agent terminal handler, gateway WebSocket bridge, control implementation, web xterm.js component) depends on the generated types being available.

This PR is intentionally minimal: only enough to keep CI green after manchtools/power-manage-sdk#25 merges. The actual feature implementation will be a separate, larger PR that:

- Adds the \`TerminalAccess\` permission and wires it through the auth system
- Implements TTY user auto-provisioning via the existing system action manager
- Mints short-lived session tokens
- Adds the gateway WebSocket endpoint and stdin audit tee
- Implements list/terminate fan-out via the new \`GatewayService\` from manchtools/power-manage-sdk#25

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet ./...\`
- [x] No existing tests change behavior — the stubs are uncalled by anything in the suite

## Refs

- manchtools/power-manage-server#6 — Remote Shell / Live Terminal (parent server feature)
- manchtools/power-manage-sdk#16 — parent SDK feature
[...]

Issue #6, opened 2026-04-03
Remote Shell / Live Terminal — interactive device access via gateway relay


PaulDotterer:
## Summary

Interactive terminal session — "access a device without opening SSH". Uses the gateway as a direct relay for low latency, with the control server handling signaling and audit only.

## Architecture

```
Web ↔ Gateway ↔ Agent     (real-time terminal I/O, direct WebSocket)
          ↓
     Control (async)       (audit log via control:inbox queue)
```

### Session lifecycle
1. Web calls `StartTerminal(deviceId, shell)` on control server
2. Control verifies permissions (`TerminalAccess` RBAC permission), creates short-lived session token, appends `TerminalSessionStarted` audit event, returns token + gateway URL
3. Web opens WebSocket to gateway: `wss://gateway/terminal?token=xxx`
4. Gateway validates token, bridges WebSocket ↔ agent stream
5. Agent opens PTY (`creack/pty`), reads/writes via the existing bidi stream
6. On disconnect: gateway appends `TerminalSessionEnded` audit event

### Audit logging
- Gateway tees **all stdin (input)** to `control:inbox` as `TerminalInputChunk` tasks (async, no latency impact)
- Captures everything: keystrokes, editor inputs (nano/vim), tab completions, ctrl sequences
- Full input replay available from the audit log
- Stdout is NOT audited (too high volume, derivable from input)

### Admin session management
- List active sessions (user, device, duration, last activity)
- Inspect live session (read-only view — admin sees terminal output but cannot type)
- Terminate session (kills PTY, closes WebSocket, audit logged)
- Permissions: `ListTerminalSessions`, `InspectTerminalSession`, `TerminateTerminalSession`

## Proto changes
- `agent.proto` — ServerMessage: `TerminalStart`, `TerminalInput`, `TerminalResize`, `TerminalStop`. AgentMessage: `TerminalOutput`, `TerminalStateChange`
- `control.proto` — RPCs: `StartTerminal`, `StopTerminal`, `ListActiveTerminalSessions`, `TerminateTerminalSession`
- `internal.proto` — RPCs: `ListGatewayTerminalSessions`, `TerminateGatewayTerminalSession`

## Also requires changes in

- `manchtools/power-manage-sdk` — Proto definitions, generated code
- `manchtools/power-manage-agent` — New `terminal` package with PTY lifecycle
- `manchtools/power-manage-web` — xterm.js terminal component, active sessions admin view

PR #35, opened 2026-04-11
feat: ControlService.StartTerminal/StopTerminal real implementation

PaulDotterer:
> ⚠️ **Stacked on #34** (terminal RPC stubs). This PR uses #34's branch as its base; **merge #34 first**, then this PR rebases onto main automatically.

## Summary

Implements step 5 of manchtools/power-manage-sdk#16 (the user-initiated open/stop half) and addresses manchtools/power-manage-server#6. Adds the four \`Remote Terminal\` permission keys, a Valkey-backed session token store, the real \`StartTerminal\`/\`StopTerminal\` handlers, audit events, and a focused test suite.

The \`List/Terminate\` admin RPCs **still return Unimplemented** because they require gateway-side session inventory and the \`GatewayService\` fan-out path; those land in the next PR alongside the gateway WebSocket bridge.

### \`internal/terminal/\` (new package)

A minimal token store layered on a tiny \`SessionBackend\` interface so handler tests can fake it without miniredis.

- **\`ValkeyBackend\`** persists tokens under \`pm:terminal:session:<id>\` with Valkey-native TTL eviction. Reuses the existing \`*redis.Client\` the control server already maintains for RediSearch — **no new dependencies, no new connection pool**.
- **\`FakeBackend\`** honours TTL via lazy expiry on read against an injectable clock so unit tests can advance time deterministically.
- **Bearer tokens** are 32 random bytes (256 bits, well above the unguessable threshold), base64url-encoded, and **SHA-256 hashed before storage**. The plaintext is returned to the web client exactly once and never persisted, so a Valkey dump cannot be used to forge connections — verified by an explicit test.
- **9 race-clean tests** cover mint/lookup round-trip, hash-not-plaintext storage, validate happy path, mismatched-token rejection (\`ErrTokenMismatch\`), unknown-session lookup (\`ErrTokenNotFound\`), idempotent revoke, TTL expiry with a frozen clock, mint required-field validation, and 100-iteration session_id uniqueness.

### \`internal/auth/permissions.go\`

Adds 4 permission keys under a new \`Remote Terminal\` group: \`StartTerminal\`, \`StopTerminal\`, \`ListActiveTerminalSessions\`, \`TerminateTerminalSession\`. The \`AuthzInterceptor\` already uses RPC method name as the permission key, so **no interceptor changes are needed**.

### \`internal/api/terminal_handler.go\` (new)

- **\`StartTerminal\`**: pulls \`UserContext\` from the request context, looks up the user, refuses disabled/deleted accounts, requires a non-empty \`linux_username\` (returns \`ErrTerminalLinuxUsernameNotSet\` / \`FailedPrecondition\` otherwise), resolves the dedicated TTY user via \`sdkterminal.TTYUsername\`, validates the target device, mints a token, emits a \`TerminalSessionStarted\` audit event, and returns the \`StartTerminalResponse\` with cols/rows/expiry/token.
- **\`StopTerminal\`**: looks up the session, **enforces ownership** (only the user that opened it may stop it — admins use \`TerminateTerminalSession\`, separate permission), revokes the token, emits a \`TerminalSessionStopped\` audit event. **Idempotent**: an unknown session returns OK with no body, matching the contract documented above \`StopTerminalRequest\` in the SDK proto.
- **\`GatewayBaseURL\`** helper normalises the configured WebSocket URL — strips any token/query/fragment defensively so the client always controls token placement. Verified by a table test that no \`?\` or \`#\` leaks through.

### \`internal/api/service.go\`

\`ControlService\` gains a \`TerminalHandler\` field (nil-by-default) and \`SetTerminalHandler\` setter, called from \`main.go\` after Valkey wiring. The 4 stubs are replaced with delegating wrappers that return:

- \`CodeUnavailable\` for \`StartTerminal\`/\`StopTerminal\` when the handler is unconfigured (so control instances without Valkey degrade gracefully instead of panicking on a nil pointer).
- \`CodeUnimplemented\` with explicit \"requires gateway fan-out\" messaging for \`ListActiveTerminalSessions\`/\`TerminateTerminalSession\`, separating \"not yet\" from \"broken\".

### \`cmd/control/main.go\`

New \`CONTROL_TERMINAL_GATEWAY_URL\` env var (and \`-terminal-gateway-url\` flag) holds the public WebSocket URL of the gateway terminal endpoint, e.g. \`wss://gw.example.com/terminal\`. **Kept separate from the existing \`CONTROL_GATEWAY_URL\`** (which is the agent registration URL) because the two endpoints run on different paths and the URLs may differ if a reverse proxy is involved.

The terminal handler is wired only when **both** Valkey is configured AND \`TerminalGatewayURL\` is set; otherwise the \`StartTerminal\` RPC returns \`CodeUnavailable\` instead of crashing on nil.

### Tests

\`internal/api/terminal_handler_test.go\` — 9 handler tests:

| Test | Asserts |
|---|---|
| \`StartTerminal_HappyPath\` | mint round-trip + token validates against the recorded session |
| \`StartTerminal_DefaultsWhenColsRowsZero\` | 80x24 default applied |
| \`StartTerminal_NoLinuxUsername\` | \`FailedPrecondition\` |
| \`StartTerminal_DeviceNotFound\` | \`NotFound\` |
| \`StartTerminal_NotAuthenticated\` | \`Unauthenticated\` |
| \`StopTerminal_OwnerCanStop\` | session removed from store |
| \`StopTerminal_OtherUserCannotStop\` | \`PermissionDenied\`, **session NOT revoked** |
| \`StopTerminal_UnknownSessionIsIdempotent\` | OK |
| \`StopTerminal_NotAuthenticated\` | \`Unauthenticated\` |
| \`GatewayBaseURL_StripsTokenAndTrailingSlash\` | no \`?\`/\`#\` leak |

All race-clean.

## Architecture decisions worth flagging

- **Opaque token + Valkey, not signed JWT.** Two reasons: (1) revocability — \`StopTerminal\` and the future admin terminate path need to invalidate tokens before expiry, which a stateless JWT can't support; (2) Valkey is already wired in for the Asynq queue and RediSearch, so adding a second backing store would be wasted complexity.

- **Permission key === RPC method name.** The interceptor enforces this convention for every other handler. The issue's spec called the permission \"TerminalAccess\" but that's a marketing term — the implementation uses the literal \`StartTerminal\` key so the existing interceptor wiring just works.

- **Separate \`TerminalGatewayURL\` config.** \`CONTROL_GATEWAY_URL\` is reserved for agent registration. Reusing it for terminal would force the gateway to expose the terminal WebSocket on the same path/scheme as the agent gRPC stream, which is wrong for most reverse-proxy setups. Two configs, two URLs, no surprises.

## What's NOT in this PR

- **Gateway WebSocket bridge** (step 4 of the issue) — needs the \`InternalService.ProxyValidateTerminalToken\` RPC in the SDK proto first, which is itself a separate small SDK PR.
- **Admin \`ListActiveTerminalSessions\` / \`TerminateTerminalSession\`** — requires the \`GatewayService\` fan-out path; lands with the gateway PR.
- **Auto-provisioning of \`pm-tty-*\` users** when \`StartTerminal\` permission is granted — separate PR via the system action manager.
- **Web xterm.js component** (step 6) — separate PR in the web repo.

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet ./...\`
- [x] \`go test ./internal/terminal/ -count=1 -race\` — all 9 store tests pass
- [x] \`go test ./internal/api/ -count=1 -run \"TestTokenStore|TestStartTerminal|TestStopTerminal|TestGatewayBaseURL\"\` — all 9 handler tests pass (78s with testcontainers)
- [x] Spot-checked a few unrelated handler tests to confirm I didn't break shared infra (TestLogin_Success, TestGetCurrentUser pass when run individually)

## Refs

- manchtools/power-manage-sdk#16 — parent feature
- manchtools/power-manage-server#6 — server-side parent issue
- #34 — base branch (terminal RPC stubs, must merge first)

## Summary by CodeRabbit

* **New Features**
  * Configurable remote terminal gateway via CLI/env; start/stop remote terminal sessions with short‑lived session tokens, persisted session store, Valkey-backed and in-memory backends, and gateway URL normalization.

* **Permissions**
  * Added remote-terminal permissions (Start, Stop, List, Terminate); StopTerminal enabled for default user role.

* **Tests**
  * End-to-end tests for terminal lifecycle, token store (mint/validate/revoke, TTL), gateway URL normalization, and authorization.

### DIFF
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/manchtools/power-manage/server/internal/scim"
 	"github.com/manchtools/power-manage/server/internal/search"
 	"github.com/manchtools/power-manage/server/internal/store"
+	"github.com/manchtools/power-manage/server/internal/terminal"
 	"github.com/manchtools/power-manage/server/internal/taskqueue"
 )
 
@@ -53,6 +54,7 @@ type Config struct {
 	AdminPassword                string
 	CORSOrigins                  []string
 	GatewayURL                   string
+	TerminalGatewayURL           string // public WebSocket URL of the gateway terminal endpoint, e.g. wss://gw.example.com/terminal
 	DynamicGroupEvalInterval     time.Duration
 	PasswordAuthEnabled          bool
 	SSOCallbackBaseURL           string
@@ -346,6 +348,26 @@ func main() {
 		searchIdx := search.New(rdb, st, aqClient, logger.With("component", "search"))
 		svc.SetSearchIndex(searchIdx)
 
+		// Wire the remote terminal session token store IF the operator
+		// has configured a public terminal gateway URL. Tokens live in
+		// Valkey under pm:terminal:session:* with a short TTL; minted
+		// by ControlService.StartTerminal and consumed by the gateway
+		// when the web client opens its WebSocket. The URL is returned
+		// to clients in StartTerminalResponse — GatewayBaseURL strips
+		// any query string defensively so the client controls token
+		// placement. When TerminalGatewayURL is empty, the terminal
+		// handler is left nil and StartTerminal returns CodeUnavailable.
+		if cfg.TerminalGatewayURL != "" {
+			terminalStore := terminal.NewTokenStore(terminal.NewValkeyBackend(rdb))
+			svc.SetTerminalHandler(api.NewTerminalHandler(
+				st,
+				terminalStore,
+				api.GatewayBaseURL(cfg.TerminalGatewayURL),
+				logger.With("component", "terminal_handler"),
+			))
+			logger.Info("remote terminal sessions enabled", "gateway_url", cfg.TerminalGatewayURL)
+		}
+
 		// Index audit events on insertion — the hook fires after every AppendEvent
 		// and enqueues the persisted row directly (no DB lookup in the search worker).
 		st.OnEventAppended = func(ctx context.Context, ev store.PersistedEvent) {
@@ -555,6 +577,7 @@ func parseFlags() *Config {
 	flag.StringVar(&cfg.AdminEmail, "admin-email", "", "Initial admin user email")
 	flag.StringVar(&cfg.AdminPassword, "admin-password", "", "Initial admin user password")
 	flag.StringVar(&cfg.GatewayURL, "gateway-url", "", "Gateway URL returned to agents during registration")
+	flag.StringVar(&cfg.TerminalGatewayURL, "terminal-gateway-url", "", "Public WebSocket URL of the gateway terminal endpoint (e.g. wss://gw.example.com/terminal). When empty, ControlService.StartTerminal returns CodeUnavailable.")
 	flag.DurationVar(&cfg.DynamicGroupEvalInterval, "dynamic-group-eval-interval", time.Hour, "Interval for evaluating dynamic groups (min 30m, max 8h, 0 to disable)")
 	flag.StringVar(&cfg.CATrustBundlePath, "ca-trust-bundle", "", "PEM file with trusted CA certificates for verification (supports CA rotation)")
 	flag.BoolVar(&cfg.TLSEnabled, "tls", false, "Enable TLS on public listener")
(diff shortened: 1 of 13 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbit review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PR #39, opened 2026-04-12
fix: fail on TOTPBackupCodeUsed and UserSessionInvalidated event errors

### LINKED ISSUE #38: refactor: fail RPCs when audit events cannot be persisted
## Summary

Most \`AppendEvent\` calls in the system are the **primary mutation** — the event IS the state change, projections derive from it, and if \`AppendEvent\` fails, the RPC fails. These are correct by construction.

However, two call sites treat a **primary-mutation \`AppendEvent\` failure as a warning**, silently dropping the state change. These are real bugs, not audit gaps.

## Real bugs (must fix)

| Call site | Event | What the warn-and-continue causes |
|---|---|---|
| \`totp_handler.go\` | \`TOTPBackupCodeUsed\` | **Double-spend**: the backup code is NOT marked as consumed, so the same one-time code works again on the next attempt. Violates the one-time guarantee. |
| \`user_group_handler.go\` | \`UserSessionInvalidated\` | **Security**: \`session_version\` is NOT bumped on the projection, so existing JWTs keep working. Admin thinks they revoked access but they didn't. |

### Fix

Both should return an error to the caller instead of logging a warning and continuing:

\`\`\`go
// totp_handler.go — TOTPBackupCodeUsed
if err := h.store.AppendEvent(ctx, ...); err != nil {
    return nil, apiErrorCtx(ctx, ErrInternal, connect.CodeInternal, "failed to consume backup code")
}

// user_group_handler.go — UserSessionInvalidated
func (h *UserGroupHandler) bumpUserSessionVersion(...) error {  // currently returns nothing
    if err := h.store.AppendEvent(ctx, ...); err != nil {
        return fmt.Errorf("failed to invalidate session: %w", err)
    }
    return nil
}
\`\`\`

The \`bumpUserSessionVersion\` signature change cascades to callers (\`bumpSessionVersionForGroupMembers\` and anywhere else that calls it), which is why this is a separate PR rather than a one-liner.

## Not bugs (warn-and-continue is fine)

| Call site | Event | Why it's OK |
|---|---|---|
| \`auth_handler.go\` | \`UserLoggedIn\` | Only drives \`last_login_at\` on the projection (cosmetic). Login works regardless — the JWT is a stateless computation from the user projection, not from this event. |
| \`sso_handler.go\` | \`UserLoggedIn\` (SSO) | Same as above |
| \`totp_handler.go\` | \`UserLoggedIn\` (TOTP) | Same as above |

These are genuinely audit-only. The state change they record (\`last_login_at\`) is nice-to-have, not security-critical. The login mutation (JWT) is not event-sourced and doesn't need to be.

## Terminal sessions (already fixed)

\`StartTerminal\`/\`StopTerminal\` were reordered in #35 to write the event first (source of truth) and the Valkey token second (derived state). Correct by construction now.

## Scope

1. Fix the 2 bugs above (small, focused PR)
2. Optionally add an \`appendRequiredEvent\` helper so future primary-mutation call sites get the right failure mode by default

## Related

- manchtools/power-manage-server#35 — where the terminal ordering was fixed
- manchtools/power-manage-sdk#16 — terminal feature that surfaced the question

PaulDotterer:
## Summary

Two bugs where \`AppendEvent\` failures were silently swallowed via warn-and-continue, causing CQRS invariant violations. Both are **primary mutations** — the event IS the state change — not secondary audit records.

Closes #38.

## Bug 1: TOTP backup code double-spend

**\`totp_handler.go\` — \`TOTPBackupCodeUsed\`**

The event marks a backup code index as consumed in the TOTP projection. If \`AppendEvent\` failed, the code was NOT consumed — the same one-time code worked again on the next attempt.

**Fix:** Return \`ErrInternal/CodeInternal\` so the caller retries rather than proceeding with an unconsumed code. The TOTP verification succeeds but the login flow fails — the user retries and the backup code gets properly consumed on the second attempt.

## Bug 2: Session invalidation silently dropped

**\`user_group_handler.go\` + \`role_handler.go\` — \`UserSessionInvalidated\`**

The event bumps \`session_version\` on the user projection, which invalidates existing JWTs. If \`AppendEvent\` failed, \`session_version\` was NOT bumped — the admin thought they revoked access but existing JWTs kept working.

**Fix:** Both the \`UserGroupHandler\` and \`RoleHandler\` copies of \`bumpUserSessionVersion\` now return \`error\`.

| Caller type | Behaviour on failure |
|---|---|
| Direct (\`AssignRoleToUser\`, \`RevokeRoleFromUser\`) | **Fail the RPC.** The role change and session invalidation must be atomic from the admin's perspective. |
| Batch (\`bumpSessionVersionForGroupMembers\`, \`bumpSessionVersionForRole\`) | **Log and continue.** Partial invalidation (some members bumped) is better than no invalidation, and the primary mutation (role/group change) is already committed. |
| Post-mutation (\`AddUserToGroup\`, \`RemoveUserFromGroup\`, \`EvaluateDynamicUserGroup\`) | **Log and continue.** The member add/remove event is already persisted; the session bump is a follow-up consistency step. |

## What was NOT a bug

The \`UserLoggedIn\` events in \`auth_handler.go\`, \`sso_handler.go\`, and \`totp_handler.go\` only drive \`last_login_at\` on the projection (cosmetic). The login mutation (JWT issuance) is a stateless computation, not event-sourced. Warn-and-continue is correct for these.

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet ./...\`
- [x] \`go test ./internal/api/ -run \"TestAssignRole|TestRevokeRole|TestAddUserToGroup|TestRemoveUserFromGroup|TestVerifyLoginTOTP\"\` — all pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Role assignment/revocation and TOTP backup-code verification now abort when the system cannot record required session or audit changes, preventing silent continuation and ensuring failures are surfaced.

* **Improvements**
  * Session invalidation for role and group operations now performs best-effort partial invalidation, reporting failures with clearer logging and propagated errors so problematic invalidation steps are visible.

### DIFF
--- a/internal/api/role_handler.go
+++ b/internal/api/role_handler.go
@@ -359,19 +359,30 @@ func (h *RoleHandler) RevokeRoleFromUser(ctx context.Context, req *connect.Reque
 		return nil, err
 	}
 
-	streamID := req.Msg.UserId + ":" + req.Msg.RoleId
-	if err := appendEvent(ctx, h.store, h.logger, store.Event{
-		StreamType: "user_role",
-		StreamID:   streamID,
-		EventType:  "UserRoleRevoked",
-		Data: map[string]any{
-			"user_id": req.Msg.UserId,
-			"role_id": req.Msg.RoleId,
-		},
-		ActorType: "user",
-		ActorID:   userCtx.ID,
-	}, "failed to revoke role"); err != nil {
-		return nil, err
+	// Check if the user currently has the role — skip the event on retry/idempotent call
+	hasRole, err := h.store.Queries().UserHasRole(ctx, db.UserHasRoleParams{
+		UserID: req.Msg.UserId,
+		RoleID: req.Msg.RoleId,
+	})
+	if err != nil {
+		return nil, apiErrorCtx(ctx, ErrInternal, connect.CodeInternal, "failed to check role assignment")
+	}
+
+	if hasRole {
+		streamID := req.Msg.UserId + ":" + req.Msg.RoleId
+		if err := appendEvent(ctx, h.store, h.logger, store.Event{
+			StreamType: "user_role",
+			StreamID:   streamID,
+			EventType:  "UserRoleRevoked",
+			Data: map[string]any{
+				"user_id": req.Msg.UserId,
+				"role_id": req.Msg.RoleId,
+			},
+			ActorType: "user",
+			ActorID:   userCtx.ID,
+		}, "failed to revoke role"); err != nil {
+			return nil, err
+		}
 	}
 
 	// Bump user's session version to invalidate cached permissions
--- a/internal/api/role_handler.go
+++ b/internal/api/role_handler.go
@@ -285,6 +285,7 @@ func (h *RoleHandler) AssignRoleToUser(ctx context.Context, req *connect.Request
 		return nil, handleGetError(ctx, err, ErrUserNotFound, "user not found")
 	}
 
+	assignedAny := false
 	for _, roleID := range roleIDs {
 		// Verify role exists
 		_, err = q.GetRoleByID(ctx, roleID)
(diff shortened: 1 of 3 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PR #40, opened 2026-04-12
feat: gateway WebSocket terminal bridge

PaulDotterer:
> ⚠️ **Stacked on #37** (gateway routing). Merge #37 first.

## Summary

Implements step 4 of manchtools/power-manage-sdk#16 — the gateway-side WebSocket endpoint that bridges web client terminal connections to the agent's bidi stream. After this PR + its prerequisite stack land, the full data path works end-to-end:

\`\`\`
Web client (xterm.js) ↔ WebSocket ↔ Gateway bridge ↔ Agent bidi stream ↔ PTY
\`\`\`

## Architecture

The gateway now has **three listeners**:

| Listener | Port | TLS | Purpose |
|---|---|---|---|
| Agent (mTLS) | \`:8080\` | mTLS (RequireAndVerifyClientCert) | Agent bidi streams, GatewayService for admin fan-out |
| Web (TLS) | \`:8443\` (configurable) | Standard TLS (no client cert) | Terminal WebSocket for web clients |
| Ops | \`:9090\` | None | Health checks |

The web listener uses the same server cert as the mTLS listener but **no ClientAuth**, so web browsers can connect via \`wss://\`. Authentication is via the session token validated against the control server's \`ProxyValidateTerminalToken\` RPC.

## What's in this PR

See the commit message for the full file-by-file breakdown. Key pieces:

- **\`TerminalSessionRegistry\`** — routes agent TerminalOutput/TerminalStateChange messages to the matching WebSocket bridge goroutine via a buffered channel. Non-blocking: full channel drops frames (acceptable for terminal output).
- **\`TerminalBridgeHandler\`** — the \`/terminal\` WebSocket handler. Validates token, verifies agent is connected, upgrades WS, sends TerminalStart, waits for STARTED, then runs two goroutines (WS→agent for input + resize, agent→WS for output). Tees stdin to the audit queue.
- **\`GatewayServiceHandler\`** — admin List/Terminate RPCs on the mTLS listener.
- **Agent message routing** — the existing \`handleAgentMessage\` no longer drops terminal messages; it routes them through the registry.
- **Audit tee** — new \`TypeTerminalAuditChunk\` task type enqueued to \`control:inbox\` for every binary (stdin) frame.

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet ./...\`
- [x] \`go test ./internal/connection/ ./internal/handler/ ./internal/taskqueue/ ./internal/gateway/... -count=1 -race\` — all existing tests pass
[...]

PR #37, opened 2026-04-12
feat: multi-gateway routing via Valkey registry

PaulDotterer:
> ⚠️ **Stacked on #36** (ProxyValidateTerminalToken handler). Merge #36 first.

## Summary

Adds the routing layer that enables terminal sessions in a multi-gateway HA deployment. Without this, the control server always returns the same static gateway URL, so a web client may land on a gateway that doesn't host the device's agent connection — and the session fails.

Each gateway now registers itself and its device connections in Valkey. The control server queries the registry at \`StartTerminal\` time and returns the URL of the **specific** gateway hosting the target device. The web client connects directly; no cross-gateway proxy, no load-balancer hacks.

## Multi-gateway flow

1. Agent connects to any gateway via the bootstrap hostname (\`gateway.example.com\`) → gateway returns HTTP 307 to its per-instance hostname (\`<ulid>.gateway.example.com\`) → agent reconnects directly.
2. Gateway writes \`pm:device:gateway:<device_id> = <gateway_id>\` to Valkey (TTL 90s, refreshed on each heartbeat).
3. Gateway publishes \`pm:gateway:terminal:<gateway_id> = wss://<id>.gateway.example.com/terminal\` (TTL 45s, background heartbeat every 15s).
4. User clicks "start terminal" → control reads the two keys → returns the resolved URL in \`StartTerminalResponse.gateway_url\` → web client connects directly to the right gateway.

## What's in this PR

See the commit message for the full breakdown. Quick summary:

| File / package | What it does |
|---|---|
| \`internal/gateway/registry/\` (new, ~250 LOC) | Registry type + ValkeyBackend + FakeBackend + 12 tests |
| \`internal/handler/agent.go\` | Attach on connect, Refresh on heartbeat, Detach on disconnect. \`BootstrapRedirectMiddleware\` returns 307 on wildcard root requests |
| \`internal/config/config.go\` | 3 new gateway env vars: \`GATEWAY_ID\`, \`GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE\`, \`GATEWAY_BOOTSTRAP_HOST\` |
| \`cmd/gateway/main.go\` | ULID generation, registry wiring, bootstrap middleware composition |
| \`internal/api/terminal_handler.go\` | \`resolveGatewayURL\`: device → gateway → URL chain; fallback to static URL when registry is nil |
| \`cmd/control/main.go\` | Instantiates the registry alongside the token store, passes both to \`NewTerminalHandler\` |
[...]

PR #36, opened 2026-04-11
feat: InternalService.ProxyValidateTerminalToken handler

PaulDotterer:
> ⚠️ **Stacked on #35** (control real impl). This PR uses #35's branch as its base; **merge #35 first**, then this PR auto-rebases onto main.

## Summary

Implements the gateway-callable validation RPC the WebSocket bridge (next PR) needs. Uses the same Valkey-backed \`terminal.TokenStore\` that \`ControlService.StartTerminal\` mints into, so:

- Tokens minted on this control instance are validatable on this instance.
- In an HA setup with multiple control replicas, tokens minted on replica A are validatable on replica B because Valkey is shared.

This is the **third-to-last** server-side terminal PR before the gateway WebSocket bridge can land.

### \`internal/api/internal_handler.go\`

- \`InternalHandler\` gains a \`terminalTokenStore\` field and a \`SetTerminalTokenStore\` setter, called from \`main.go\` alongside \`ControlService.SetTerminalHandler\` so the two paths share one store.
- nil-safe: when terminal sessions are not configured on this control instance, \`ProxyValidateTerminalToken\` returns \`Unavailable\` so the gateway gets a clean error instead of the default \"method not implemented\".

**Handler behaviour:**

| Input | Result |
|---|---|
| empty \`session_id\` or \`token\` | \`InvalidArgument\` |
| valid token | \`OK\` + \`user_id\`/\`device_id\`/\`tty_user\`/\`cols\`/\`rows\` |
| unknown session (\`ErrTokenNotFound\`) | \`Unauthenticated\` (logged at \`Debug\`) |
| mismatched token (\`ErrTokenMismatch\`) | \`Unauthenticated\` (logged at \`Warn\` — possible forgery) |
| store not configured (nil) | \`Unavailable\` |
| unexpected store error | \`Internal\` (never leaked as \`Unauthenticated\`) |

**Forgery probe protection**: \`ErrTokenMismatch\` and \`ErrTokenNotFound\` both surface as \`Unauthenticated\` with the same generic message (\"invalid or expired session token\") so an attacker cannot distinguish expired from never-existed by error response alone. They are logged differently though, so operators can spot active attacks in audit.

**Validation does NOT consume the entry** — the contract documented on the RPC in manchtools/power-manage-sdk#27 specifies that the same gateway uses the metadata across the lifetime of the WebSocket; revocation is explicit via \`ControlService.StopTerminal\` or \`TerminateTerminalSession\`. There's a test specifically for this.

[...]

PR #41, opened 2026-04-12
feat: auto-provision pm-tty-* user accounts for terminal sessions

PaulDotterer:
## Summary

Closes the last functional gap that would cause terminal sessions to fail in production: the agent's `OnTerminalStart` requires the `pm-tty-*` account to exist on the device before it can allocate a PTY. Without this PR, every `StartTerminal` call that successfully mints a token and reaches the agent fails with "tty user not provisioned."

This PR hooks into the existing `SystemActionManager` (which already handles user provisioning and SSH access actions) to auto-create a dedicated `pm-tty-<linux_username>` User action when the PM user has `StartTerminal` permission.

Independent of the terminal PR stack (#35 → #36 → #37 → #40); lands directly on main.

## How it works

**Trigger:** `SyncUserSystemActions` is called on user create, update, role change, linux username change, etc. It now queries the user's effective permissions via `GetUserPermissionsWithGroups` and checks for `StartTerminal`.

**If the user has `StartTerminal` permission:**
- Creates a system `ACTION_TYPE_USER` action named `system:tty-user:<user_id>` with:
  - `username`: `pm-tty-<linux_username>`
  - `uid`: `linux_uid + 100000` (deterministic, matches `terminal.DefaultUIDOffset`)
  - `shell`: `/usr/sbin/nologin` (agent activates during session only)
  - `createHome`: false (agent creates per-session temp dir)
  - `system`: true (hidden from login screens via AccountsService)
  - `disabled`: mirrors the PM user's disabled state
- Assigns the action to the user (so it resolves onto the user's assigned devices via the existing assignment/resolution engine)
- Links via `system_tty_action_id` on the user projection

**If the user does NOT have `StartTerminal` permission:**
- Deletes the existing action and unlinks it

**On user deletion:** `CleanupDeletedUserActions` now also cleans up the TTY action.

## Migration 007

- Adds `system_tty_action_id TEXT NOT NULL DEFAULT ''` to `users_projection`
- Extends the `UserSystemActionLinked` case in `project_user_event` to handle the new field alongside `system_user_action_id` and `system_ssh_action_id`
- Full `CREATE OR REPLACE` of `project_user_event` because PL/pgSQL doesn't support partial function updates; body reproduces existing logic from 002 with the third `CASE` arm added

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] sqlc regeneration adds `SystemTtyActionID` to `UsersProjection` model
- [x] Migration SQL uses `+goose StatementBegin`/`End` for the PL/pgSQL function (per the project's Goose migration memory)
- [ ] Integration tests (will run in CI against a real PostgreSQL instance)

## Refs

- manchtools/power-manage-sdk#16 — parent feature
- manchtools/power-manage-server#6 — server-side parent issue

## Summary by CodeRabbit

* **New Features**
  * Automatic terminal (TTY) access is provisioned for eligible users and kept up to date.
  * System-managed terminal accounts use deterministic, consistent naming.

* **Bug Fixes**
  * Terminal access and linked accounts are removed reliably when permissions are revoked to prevent stale entries.

* **Chores**
  * Backend now tracks per-user terminal access status for auditing and synchronization.

* **Documentation**
  * Added Remote Terminal permissions and gateway/terminal configuration guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ See the [Control Server README](cmd/control/) for details on the event model, AP
 | `internal/search` | Full-text search indexer using Valkey RediSearch — FT index management, Asynq reindex workers, cascade updates |
 | `internal/store` | PostgreSQL event store, migrations, sqlc queries |
 | `internal/taskqueue` | Asynq task queue client, task type constants, payload structs |
+| `internal/terminal` | Session token store (Valkey-backed) for remote terminal sessions |
+| `internal/gateway/registry` | Multi-gateway device→gateway routing registry (Valkey-backed) |
 | `internal/testutil` | Test helpers — PostgreSQL testcontainers, test entity factories, auth context injection |
 
 ## API Reference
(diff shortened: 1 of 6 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PR #45, opened 2026-04-14
fix: re-sign action payload with execution ID in dispatchPendingActions

PaulDotterer:
## Summary
- `dispatchPendingActions` was sending the action's original signature (signed over the action ID), but the gateway sets `Action.Id = executionID` — causing agent signature verification failures on reconnect dispatch
- Now re-signs with the execution ID before enqueuing, mirroring the API dispatch handler
- Adds integration test with miniredis verifying the signer receives the execution ID, not the action ID

## Root cause
The gateway always sets `Action.Id = payload.ExecutionID` in `task_handlers.go:69`. The API dispatch path re-signs with the execution ID before enqueuing. But `dispatchPendingActions` (triggered on device hello for pending/dispatched executions) used the action's original signature — signed over the action ID. When the agent verified the signature against `Action.Id` (the execution ID), it failed.

This was a latent bug that became visible when:
1. The Asynq task from the original API dispatch was consumed but the agent disconnected before executing
2. On reconnect, `dispatchPendingActions` re-dispatched with the mismatched signature

## Test plan
- [x] `TestDispatchPendingActions_ReSignsWithExecutionID` — verifies signer is called with execution ID
- [x] All existing inbox worker tests pass
- [ ] Deploy and verify agents no longer get signature errors on reconnect dispatch

Refs manchtools/power-manage-server#43

## Summary by CodeRabbit

* **Internal Improvements**
  - Pending actions are now re-signed at dispatch via an injected signer; executions that cannot be re-signed or resolved are skipped and emit structured logs.

* **Tests**
  - Added a concurrency-safe test double and an integration-style test that verifies re-signing is invoked with the expected identifiers and canonical parameters.

* **Chores**
  - Updated indirect module requirements to support the test infrastructure.

### DIFF
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -457,7 +457,7 @@ func main() {
 		}
 
 		// Build Asynq mux with inbox worker only (search worker runs in indexer binary)
-		inboxWorker := control.NewInboxWorker(st, aqClient, logger.With("component", "inbox_worker"))
+		inboxWorker := control.NewInboxWorker(st, aqClient, actionSigner, logger.With("component", "inbox_worker"))
 		mux := inboxWorker.NewMux()
 
 		// Start Asynq server consuming control:inbox queue only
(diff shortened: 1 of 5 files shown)


PaulDotterer:
@coderabbitai full review

PaulDotterer:
@Coderabbitai full review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PR #50, opened 2026-04-17
refactor: graceful shutdown, retry backoff, remove dead fingerprint helper

PaulDotterer:
## Summary

Three independent cleanup commits surfaced while reviewing the repo for bugs, oversights, and simplification opportunities.

### 1. Graceful shutdown on listener failure (`f45527d`)

- `EnsureIndexes` at control startup was called with `context.Background()` instead of the shutdown-scoped ctx, so a hanging Redis would block graceful shutdown.
- The gateway's mTLS and web-TLS server goroutines called `os.Exit(1)` on listener failure, bypassing the in-flight Asynq / WebSocket drain that the `defer` chain otherwise runs. Switched both to `stop()` so shutdown follows the same path as the control server.

### 2. Exponential backoff between AppendEvent retries (`c95a751`)

The dispatch-event retry loop in `InboxWorker` spun three times immediately on failure, hammering the DB with no chance for a transient error to clear. Back off 100ms then 200ms between attempts, and cancel on ctx shutdown.

### 3. Remove dead `CertificateFingerprint` helper (`5c1a300`)

The helper in `internal/mtls/mtls.go` was unused in production (only its own tests called it) and the doc comment lied — it returned hex-encoded raw DER bytes, not a SHA256 hash. The real fingerprint lives in `ca.FingerprintFromPEM` and correctly computes `sha256.Sum256` over the DER bytes. Keeping the misnamed helper around is a footgun: a future caller copying it would compare hex-DER against the DB's SHA256 column and every device would fail renewal.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/control/...`
- [x] `go test ./internal/mtls/`
- [ ] Manual: kill mTLS port during gateway startup → verify orderly shutdown

## Release Notes

* **Bug Fixes**
  * Improved error handling in gateway server initialization to initiate graceful shutdown instead of abrupt termination when TLS server startup fails.
  * Added exponential backoff retry logic for event persistence operations to improve reliability during transient failures.

* **Improvements**
  * Enhanced context propagation for proper cancellation handling during server initialization and shutdown sequences.

* **Refactor**
  * Internal code restructuring for improved dependency management and maintainability.
  * Removed unused certificate fingerprinting utility.

### DIFF
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -452,7 +452,7 @@ func main() {
 		}
 
 		// Ensure indexes exist (idempotent, needed for FT.SEARCH queries).
-		if err := searchIdx.EnsureIndexes(context.Background()); err != nil {
+		if err := searchIdx.EnsureIndexes(ctx); err != nil {
 			logger.Warn("failed to ensure search indexes", "error", err)
 		}
 
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -337,7 +337,7 @@ func main() {
 		logger.Info("starting gateway server with mTLS", "address", cfg.ListenAddr)
 		if err := server.ListenAndServeTLS("", ""); err != nil && err != http.ErrServerClosed {
 			logger.Error("server error", "error", err)
-			os.Exit(1)
+			stop()
 		}
 	}()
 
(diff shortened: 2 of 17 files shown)


PaulDotterer:
@coderabbitai review

Issue #53, opened 2026-04-17
hardening: tighter SCIM rate limiting + bcrypt cost parity with password flow


PaulDotterer:
## Summary

Follow-up from the server audit. Two related SCIM/auth hardening items surfaced; neither is a live vulnerability, so tracking them here rather than on an immediate PR.

## 1. SCIM rate limiting is per-slug, not per-token

**Where:** `internal/scim/auth.go` in \`withAuth\` — \`h.rateLimiter.Allow(slug)\` gates at the provider level.

**Context:** On a shared provider (e.g. one IdP, many SCIM tokens rotated over time) every token counts against the same per-slug bucket. An attacker distributing across many IPs can't fully bypass the limit, but they can stretch it, and the signal-to-noise for rate-limit alerts is poor (can't tell which token is being brute-forced).

**Proposal:** Rate limit per \`(slug, token_hash)\` tuple. Key the limiter on \`slug + ":" + sha256(token)\` so different tokens hit independent buckets, and legitimate high-volume tokens aren't throttled by attack traffic on a sibling token.

## 2. Backup-code bcrypt cost (12) is lower than password bcrypt cost (14)

**Where:** \`internal/auth/totp/backup.go:17\` vs the password flow's cost.

**Context:** Backup codes are 64-bit random hex — even cost 12 takes ~300ms/guess on commodity GPU, so the gap isn't catastrophic. But compromised DB snapshots are cracked 4× faster against the weaker cost.

**Proposal:** Bump backup-code bcrypt cost to match the password flow (14). Benchmark latency on the hot path — verifying a single backup code on login is a one-shot operation, so even 500ms is acceptable UX-wise.

## Acceptance

- [ ] SCIM \`RateLimiter.Allow(...)\` keyed on token hash
- [ ] Backup-code cost constant raised to 14 and benchmarked
- [ ] Unit test that two different tokens for the same provider get independent rate-limit buckets

## Audit reference

Flagged during the Tier 2 pass of the server audit in this repo. All other Tier 2 claims (SCIM cross-provider authz bypass, status-code slug enumeration, backup-code timing leak, TOTP timing, Redis-pipeline silencing) were verified as non-issues.

PR #54, opened 2026-04-17
fix: server audit Tier 1 — security, concurrency, shutdown correctness

PaulDotterer:
## Summary

Tier 1 findings from the CodeRabbit-style server audit. Split into three themed commits for reviewability; every fix is independent and any one can revert cleanly.

Tier 2 validation notes and a follow-up hardening issue are tracked at #53.

## Commit 1 — security hardening (\`47ef1d1\`)

- **registration_handler**: reject one-time tokens with CurrentUses > 0 before CSR signing. Defence-in-depth; the existing \`token.Disabled\` + event-store optimistic lock already prevent real reuse, but this fails fast in the projection-lag window.
- **ca**: reject CSRs requesting Subject Alternative Names. Agent certs are CN-identified — copying \`csr.DNSNames\` verbatim into the issued cert let a malicious agent request SANs for internal hostnames that downstream verifiers might trust.
- **device_group / user_group handlers**: cap \`dynamic_query\` at 10 000 chars before validation. Untrusted input that ends up on the event stream is now length-bounded.
- **search_handler**: cap pagination offset at 100 000 instead of accepting any integer from \`strconv.Atoi\`.

## Commit 2 — concurrency and lifecycle (\`2e18f72\`)

- **connection.Agent.Send**: acquire \`sendMu\` before checking \`ctx.Done()\` — prior order had a TOCTOU window where \`Close()\` could cancel ctx between the check and the lock.
- **connection.Manager.Register**: derive agent ctx from a parent ctx argument (the handler's RPC ctx), not \`context.Background()\`. Agent lifetime now ends when the handler goroutine does. Signature change ripples to \`handler/agent.go\` and the unit tests.
- **handler/terminal_bridge.bridgeAgentToWS**: add a \`default\` branch to the type switch. Unknown agent message types are now logged rather than silently dropped.
- **control/inbox_worker.dispatchPendingActions**: mark execution failed via \`ExecutionFailed\` event when \`GetActionByID\` returns \`pgx.ErrNoRows\`. Fixes the "stuck pending forever" loop that re-ran the same orphaned execution on every reconnect.

## Commit 3 — startup and shutdown (\`15244c3\`)

- **cmd/control**: raise public-listener \`MinVersion\` from TLS 1.2 to 1.3, matching the internal mTLS listener.
- **cmd/control**: fail closed when \`PM_ENCRYPTION_KEY\` is unset. Operators who genuinely want plaintext secret storage can opt in with \`PM_ENCRYPTION_KEY_REQUIRED=false\`.
- **cmd/control**: rewrite \`maskDatabaseURL\` with \`net/url.Parse\` rather than a hand-rolled scan that mangled URL-encoded passwords.
- **cmd/gateway**: declare \`shutdownCtx\` early and derive the internal-URL refresh goroutine from it. Previously that goroutine ticked past SIGTERM until the function-exit defer ran.

## Audit findings ruled out

Several subagent-flagged items were verified as false positives against the current code:

- \`RebuildSearchIndex\` "missing auth": the \`AuthzInterceptor\` chain at \`cmd/control/main.go:505\` already gates every non-public RPC against \`HasPermission(procedureName)\`, and \`RebuildSearchIndex\` is a registered permission.
- OIDC \`redirect_uri\` whitelist: the \`auth_states\` row captures the exact URI at login time and the callback re-reads it; combined with IdP-side pre-registration this is already the standard mitigation.
- SCIM cross-provider authz bypass: \`bcrypt.CompareHashAndPassword\` at \`scim/auth.go:79\` binds the token to the slug's stored hash.
- SCIM slug enumeration via status codes: all failure paths return 401.
- TOTP / backup-code timing: \`pquerna/otp\` and \`bcrypt\` are constant-time; loop-index leak only reveals which of the attacker's own codes matched.
- Redis pipeline error silencing in search-index warm: intentional best-effort; reconciliation restores state.

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet ./...\`
- [x] \`go test -short ./internal/ca/ ./internal/connection/ ./internal/mtls/ ./internal/auth/totp/\`
- [ ] Integration: re-run testcontainers suite on CI
- [ ] Manual: start control with PM_ENCRYPTION_KEY unset — verify fail-closed behaviour
- [ ] Manual: verify \`GATEWAY_ID\`-enabled gateway shuts down cleanly on SIGTERM without ticking refresh logs

## Follow-up

- #53 — SCIM rate limiting per-token + bcrypt cost parity (backlog, not blocking)

### DIFF
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	urlpkg "net/url"
 	"os"
 	"os/signal"
 	"strings"
@@ -272,7 +273,17 @@ func main() {
 		os.Exit(1)
 	}
 	if encryptor == nil {
-		logger.Warn("PM_ENCRYPTION_KEY not set - secrets will be stored unencrypted")
+		// Fail closed: IdP client secrets, LUKS keys, and other
+		// secrets-at-rest rely on this encryptor. Running without a
+		// key would silently degrade security in production. Operators
+		// who truly want unencrypted storage can set
+		// PM_ENCRYPTION_KEY_REQUIRED=false to opt in explicitly.
+		if os.Getenv("PM_ENCRYPTION_KEY_REQUIRED") == "false" {
+			logger.Warn("PM_ENCRYPTION_KEY not set and PM_ENCRYPTION_KEY_REQUIRED=false - secrets will be stored unencrypted")
+		} else {
+			logger.Error("PM_ENCRYPTION_KEY is required (set PM_ENCRYPTION_KEY_REQUIRED=false to opt out)")
+			os.Exit(1)
+		}
 	}
 
 	// Initialize action signer (signs actions so agents can verify authenticity)
@@ -533,7 +544,7 @@ func main() {
 		}
 		server.TLSConfig = &tls.Config{
 			Certificates: []tls.Certificate{cert},
-			MinVersion:   tls.VersionTLS12,
+			MinVersion:   tls.VersionTLS13,
 		}
 		if err := http2.ConfigureServer(server, &http2.Server{}); err != nil {
 			logger.Error("failed to configure HTTP/2 for public server", "error", err)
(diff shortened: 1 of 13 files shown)


PaulDotterer:
@coderabbitai review

PR #64, opened 2026-04-19
fix: gateway terminal listener cleartext behind Traefik (rc4)

PaulDotterer:
## Summary

Fixes the terminal WebSocket 400 exposed once rc3 got past the Let's Encrypt cert.

The gateway's `/terminal` listener was calling `ListenAndServeTLS` with the internally-signed control cert. Traefik's TTY router forwards **cleartext HTTP** to it (the `ttyBackend` auto-derives as `http://<hostname>:<port>` and that's the documented shape), so every browser upgrade round-tripped into Go's \`tls.Conn\`:

\`\`\`
HTTP/2 400
Client sent an HTTP request to an HTTPS server.
\`\`\`

This was a latent rc2 bug hidden by the `/tls=\"true\"` cert regression fixed in rc3 — once browsers stopped bouncing off the self-signed cert, the backend-protocol mismatch became the new failure.

## Change

Strip TLS from the gateway's web listener entirely. Public TLS termination stays at Traefik (Let's Encrypt via the per-replica certResolver); Traefik forwards cleartext to the gateway over the internal `pm-internal` docker/k8s network — the standard Traefik-behind-service pattern.

- Drops a crypto layer that served no real purpose (the cert was internally-signed and no browser client ever saw it directly).
- Fixes the WebSocket upgrade path.
- mTLS listener on `:8080` is **unchanged** — agents still get the full client-cert handshake there.
- Ops/health listener unchanged.

## Test plan

- [x] \`go build ./...\` / \`go vet ./...\` clean
- [x] \`go test ./internal/gateway/... ./internal/handler/...\`
- [ ] Operator verification: after rc4 deploy, \`curl -i https://\${TTY_DOMAIN}/gw/\${GATEWAY_ID}/terminal\` should return \`400 session_id and token query parameters are required\` (Go handler reached, not TLS error), and the browser terminal should open.

## Notes

No env-var or compose changes needed. The existing reference `compose.yml` already ships `ttyBackend` as `http://`. Operators with custom `GATEWAY_TRAEFIK_TTY_BACKEND` values need to make sure theirs is `http://`, not `https://`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified gateway docs to state the TTY WebSocket listener is cleartext HTTP with TLS terminated upstream.

* **Configuration**
  * WebSocket listener now accepts cleartext HTTP while TLS is handled upstream.

* **Validation**
  * TTY backend addresses are now validated to require an http scheme and a non-empty host.

* **Tests**
  * Added tests covering rejection of invalid TTY backend URLs.

### DIFF
--- a/cmd/gateway/README.md
+++ b/cmd/gateway/README.md
@@ -60,7 +60,7 @@ The Gateway Server:
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `GATEWAY_LISTEN_ADDR` | `:8080` | Listen address for the agent mTLS listener |
-| `GATEWAY_WEB_LISTEN_ADDR` | (empty) | Listen address for the TTY WebSocket listener (empty disables the terminal feature) |
+| `GATEWAY_WEB_LISTEN_ADDR` | (empty) | Listen address for the TTY WebSocket listener (cleartext HTTP — public TLS is terminated at Traefik; empty disables the terminal feature) |
 | `GATEWAY_VALKEY_ADDR` | `localhost:6379` | Valkey/Redis address for Asynq task queue |
 | `GATEWAY_VALKEY_PASSWORD` | (empty) | Valkey/Redis password |
 | `GATEWAY_VALKEY_DB` | `0` | Valkey/Redis database number |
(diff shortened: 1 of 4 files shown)


PaulDotterer:
@coderabbitai full review

PR #66, opened 2026-04-20
feat: rc6 gateway stability polish + audit chunk coalescing

PaulDotterer:
## Summary

Four stability/DX improvements to the terminal path. Server-only, no proto or migration changes.

## 1. Auto-derive \`GATEWAY_INTERNAL_URL\`

The admin RPCs for listing/terminating terminal sessions fan out from control to every gateway's internal mTLS listener. That listener is the same one agents use, so the internal URL is just \`https://<hostname><mtls-port>\` — yet it was only published to the registry when operators set \`GATEWAY_INTERNAL_URL\` explicitly. Common-case deployments left it empty, control's fan-out found zero gateways, and the Terminal Sessions page silently rendered an empty list.

**Fix:** auto-derive when unset, matching the pattern we already use for \`MTLSBackend\` / \`TTYBackend\`. Zero-config now produces a working admin list.

## 2. Split WebSocket close codes on the terminal bridge

All three failure paths in \`waitForStarted\` used \`StatusInternalError\` (1011), so browsers couldn't tell "agent refused by policy" (retry pointless) from "transient stall" (retry likely works) from "genuine bug." Split into:

- \`StatusPolicyViolation\` (1008) — agent refused the session (e.g. \`tty.enabled=false\`, missing \`pm-tty-*\` user)
- \`StatusTryAgainLater\` (1013) — agent start timeout (transient)
- \`StatusInternalError\` (1011) — retained for "channel closed" / "exited prematurely" paths

Reason text still carries the detail; the close code is now the machine-readable signal.

## 3. Coalesce audit chunks (4 KiB / 1 s)

**Before:** xterm.js sends one WebSocket frame per keystroke, so the audit tee produced **one \`TerminalInputChunk\` event per character**. A typed command became 10+ rows of opaque single-byte base64 blobs in the event store, and an active session flooded the inbox queue.

**After:** \`terminalAuditBatcher\` buffers stdin up to **4 KiB** or **1 s of typing pause** (true debounce — each keystroke resets the timer), then emits one chunk with a monotonic sequence. Typing path is **untouched** — the batcher is a parallel in-memory tee. \`Close\` flushes pending so a clean disconnect loses no bytes; a hard gateway crash loses at most ~1 s of keystrokes (~100 bytes typical).

This is a narrow interim fix. The long-term reshape — moving session data off the audit stream into a dedicated \`terminal_sessions\` table — is tracked as a separate issue for rc7.

## 4. Log registered terminal mux routes at startup

rc3-rc5 each stumbled on a mismatch between the minted \`PublicTerminalURLTemplate\` and the handler the gateway actually registered. Emit the full list of handler paths at startup so operators can eyeball them against Traefik rules without curl-probing:

\`\`\`
INFO terminal mux routes registered paths=[/terminal /gw/01KP.../terminal] health_path=/health
\`\`\`

## Test plan

- [x] \`go build ./...\` / \`go vet ./...\` clean
- [x] \`go test -race -count=1 ./internal/handler/...\` green
- [x] 7 new batcher tests (size-cap flush, keystroke coalescing, monotonic seq across flushes, Close flushes pending, Close idempotent, write-after-close no-op, empty write no-op)
- [ ] Operator verification: Terminal Sessions page shows live sessions; closed-session audit events are one-per-typed-line, not one-per-keystroke

## Pairs with

- manchtools/power-manage-agent#49 — grants CAP_SETUID / CAP_SETGID so pty spawn stops hitting EPERM. Both land together in the 2026.05-rc6 + agent 2026.05-rc3 cut.

## Summary by CodeRabbit

* **Bug Fixes**
  * More appropriate WebSocket close codes for terminal connections to better reflect different failure scenarios.

* **Improvements**
  * Gateway startup now auto-derives and logs its internal URL and concrete terminal/health routes for clearer diagnostics.
  * Terminal audit logging now buffers and batches input to reduce overhead while preserving ordering and ensuring pending data is flushed on session end.

* **Tests**
  * Added comprehensive tests covering terminal audit batching behavior.

### DIFF
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -198,9 +198,27 @@ func main() {
 		// Also publish the internal mTLS URL so the control server
 		// can discover this gateway for admin fan-out (List/Terminate
 		// terminal sessions). Uses the same TTL as the terminal URL.
-		if cfg.InternalURL != "" {
+		//
+		// rc6 note: auto-derive when GATEWAY_INTERNAL_URL is empty.
+		// The GatewayService RPC is mounted on the same mTLS listener
+		// that accepts agents, so the internal URL is just
+		// `https://<hostname><listen-port>`. Auto-derivation keeps the
+		// common case zero-config — without it, operators who don't
+		// set the env silently lose the admin list/terminate feature
+		// (Terminal Sessions page shows empty).
+		internalURL := cfg.InternalURL
+		if internalURL == "" {
+			hostname, err := os.Hostname()
+			if err != nil || hostname == "" {
+				logger.Error("cannot auto-derive GATEWAY_INTERNAL_URL: os.Hostname failed", "error", err)
+				os.Exit(1)
+			}
+			internalURL = "https://" + hostname + portOfListenAddr(cfg.ListenAddr)
+			logger.Info("auto-derived GATEWAY_INTERNAL_URL", "internal_url", internalURL)
+		}
+		if internalURL != "" {
 			if err := gatewayReg.RegisterGatewayInternal(
-				context.Background(), gatewayID, cfg.InternalURL, registry.DefaultGatewayTTL,
+				context.Background(), gatewayID, internalURL, registry.DefaultGatewayTTL,
 			); err != nil {
 				logger.Warn("failed to register gateway internal URL", "error", err)
 			}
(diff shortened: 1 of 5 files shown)


PaulDotterer:
@coderabbitai full review

PaulDotterer:
@coderabbitai full review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PR #72, opened 2026-04-21
fix: server audit pass (peer-class, inline validation, fail-closed signing, key perms)

PaulDotterer:
## Summary

Addresses the server-sweep audit findings plus related code smells. All changes are within the no-backwards-compat window.

### High severity

**Private key permissions (setup.sh)**. Every private key is now chmod 0600 and the script runs under `umask 077` so openssl never briefly writes the key wider. The CA key particularly was world-readable before.

**Peer-class mTLS**. New \`internal/mtls/peer_class.go\` defines a SPIFFE URI SAN (\`spiffe://power-manage/<class>\`) stamped onto every cert. CA now emits \"agent\" on issued certs; setup.sh stamps \"gateway\" / \"control\" on its out-of-band certs. \`RequirePeerClass\` middleware enforces the expected class on each listener:

- control \`InternalService\` admits only \`gateway\`
- gateway \`AgentService\` admits only \`agent\`
- gateway \`GatewayService\` admits only \`control\`

A leaked agent cert can no longer be replayed against the internal listener to read other devices' LUKS/LPS secrets; a leaked agent cert can no longer pose as the control plane to issue admin fan-out RPCs.

### Medium severity

**Inline dispatch validation**. \`DispatchAction\`'s inline branch now runs the same per-oneof validation Create applies, including the shell \"at least one of script or detection_script\" rule. Previously an inline dispatch could smuggle a malformed or nil payload past Create's gate; nil \`inline_action\` even panicked \`extractActionParamsMsg\`. \`validateInlineActionPayload\` rejects nil before downstream serialization runs.

**\`signAction\` fail-closed**. Signing or persistence failure now aborts the Create / Update / Dispatch request with \`CodeInternal\` instead of being logged-and-swallowed. Previously the server could record an unsigned action that the agent silently drops on dispatch, leaving the execution in \"pending forever\" state with no clue to the operator. The Dispatch re-sign path also joins this policy.

**Update shell parity**. \`UpdateActionParams\` with a Shell payload now runs the \`Script | DetectionScript\` required-oneof check that Create has always done. Editing a valid shell action to blank both scripts is now rejected instead of signing a no-op.

### Low/medium — terminal bearer

Accept session token via \`Sec-WebSocket-Protocol: bearer.<token>\` as the preferred transport. Query-param \`?token=\` still works for legacy clients but logs a WARN naming the remote address so operators can trace the stragglers. Moving the token off the URL reduces exposure via proxy access logs, referrer headers, and browser devtools.

### Code smells

- Migrate from the deprecated \`nhooyr.io/websocket\` to its active successor \`github.com/coder/websocket\`. Drop-in rename; no API differences in what we use.
- Drop unused \`executionWarmDays\` / \`auditWarmDays\` constants in \`internal/search/index.go\`.
- README cookie-auth drift: bearer-only is the actual transport; the README's cookie section described a path that no longer exists in code. Also documents peer-class enforcement in the mTLS section.

## Migration

Operators upgrading need to:
1. Re-run \`server/deploy/setup.sh\` with \`--force\` to regenerate gateway + control certs with the new URI SANs.
2. Agents with pre-upgrade certs (no URI SAN) cannot reach the gateway until they renew. They CAN reach \`ControlService.RenewCertificate\` (which is on the public, LE-fronted listener — no peer-class enforcement there). CA-rotation trust-bundle mechanism triggers fleet-wide renewal.

## Test plan

- [x] \`go build ./...\` — clean
- [x] \`go vet ./...\` — clean
- [x] \`go test ./internal/auth/... ./internal/ca/... ./internal/handler/... ./internal/gateway/... ./internal/connection/... ./internal/terminal/... ./internal/taskqueue/... ./internal/config/... ./internal/mtls/... ./internal/search/... ./internal/middleware/...\` — all green
- [ ] \`go test ./internal/api/...\` — the long integration suite hit the local 900s timeout; CI should complete it
- New: \`internal/mtls/peer_class_test.go\` — roundtrip, rejection surface, middleware allow/deny, health bypass, missing-TLS 401
- New: \`internal/api/dispatch_validation_test.go\` — inline nil rejection, inline shell-missing-script rejection, update shell-missing-script rejection

## Summary by CodeRabbit

* **New Features**
  * mTLS now enforces peer-class roles (agent, gateway, control) across listeners and rejects mismatched peers.
  * Device certificates and server certificates include SPIFFE peer-class URI SANs.

* **Bug Fixes**
  * Dispatch and signing paths are fail-closed; compensating events recorded on enqueue/sign failures.
  * Handlers return errors when no task-queue is configured to avoid dangling pending state.

* **Security Improvements**
  * JWTs moved to bearer-only transport; server-issued auth cookies removed.

* **Tests**
  * Added and updated tests and test helpers for peer-class, validation, enqueue, and signing behaviors.

* **Chores**
  * WebSocket library updated; script/file permissions tightened.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ See the [Control Server README](cmd/control/) for details on the event model, AP
 | Package | Purpose |
 |---------|---------|
 | `internal/api` | Control Server RPC handlers (actions, devices, users, tokens, assignments, roles, user groups, identity providers, SCIM, TOTP, compliance, etc.) |
-| `internal/auth` | JWT authentication, OPA authorization, TOTP 2FA, rate limiting, cookie management, self-scope enforcement |
+| `internal/auth` | JWT bearer authentication, OPA authorization, TOTP 2FA, rate limiting, self-scope enforcement |
 | `internal/ca` | Internal CA for signing agent certificates, certificate renewal verification, action payloads, and CA rotation via trust bundles |
 | `internal/config` | Configuration loading (gateway) |
 | `internal/connection` | Gateway connection manager — tracks connected agents, routes messages |
@@ -66,7 +66,7 @@ See the [Control Server README](cmd/control/) for details on the event model, AP
 | `internal/handler` | Gateway RPC handlers (agent streaming, Connect-RPC proxy to control, auto-update info) |
 | `internal/idp` | OIDC identity provider SSO (authorization code flow, token exchange, user linking) |
 | `internal/middleware` | HTTP middleware (request ID injection, security headers, logging) |
-| `internal/mtls` | mTLS setup (`RequireAndVerifyClientCert`, TLS 1.3), extracts device identity from client certificates |
+| `internal/mtls` | mTLS setup (`RequireAndVerifyClientCert`, TLS 1.3), extracts device identity from client certs, enforces SPIFFE peer-class URI SANs (`agent` / `gateway` / `control`) on each listener |
 | `internal/resolution` | Assignment resolution engine (user/user_group/device/device_group targets) |
 | `internal/scim` | SCIM v2 provisioning server (REST endpoints for user/group sync from external IdPs) |
 | `internal/search` | Full-text search indexer using Valkey RediSearch — FT index management, Asynq reindex workers, cascade updates |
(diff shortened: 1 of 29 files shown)


PaulDotterer:
@coderabbitai review

PR #76, opened 2026-04-23
fix: rc10 consolidation — auth bypass, fail-open gateway, deploy hardening, projections, data integrity

### LINKED ISSUE #74: fix: gateway Traefik backend auto-derivation publishes unresolvable container ID
## Problem

When `GATEWAY_TRAEFIK_SELF_REGISTER=true`, each gateway replica auto-derives its own `mtlsBackend` / `ttyBackend` from `os.Hostname()` and publishes it to Traefik's Redis KV provider. Inside a Docker container the default hostname is the 12-char container ID (e.g. `bb3ac39ada41`), which is **not registered in Docker's embedded DNS** for user-defined networks.

Result: Traefik sees a backend like `bb3ac39ada41:8080`, fails to resolve it, and the pm-mtls TCP router silently loses its healthy backend. Agent mTLS traffic then falls through to the HTTP router on the same `:443` entrypoint, which terminates with the Let's Encrypt cert — the agent sees `x509: certificate signed by unknown authority` and can never connect.

Observed on staging during the rc7/rc8 cycle.

## Secondary problem: env-var explosion

Seven `GATEWAY_TRAEFIK_*` env vars had to be wired through compose.yml by hand, every real deployment set them to the same values, and a typo in any one of them (e.g. `TTY_DOMAIN` vs `GATEWAY_TTY_DOMAIN`) breaks routing silently.

## Fix

- Replace `os.Hostname()` with a `routableIP()` helper that returns the replica's own non-loopback IPv4 on the shared container network (via `net.InterfaceAddrs()`). Used for both Traefik backends and the auto-derived `GATEWAY_INTERNAL_URL`.
- Move the Traefik defaults into the gateway binary:
  - `SelfRegister=true`, `RootKey=traefik`, `MTLSEntryPoint=websecure`, `TTYEntryPoint=websecure`, `TTYCertResolver=letsencrypt`
  - `MTLSHost` falls back to `GATEWAY_DOMAIN`
  - `TTYHost` falls back to `GATEWAY_TTY_DOMAIN`
- Strip the matching 7 env lines from `deploy/compose.yml` (operators now only need `GATEWAY_DOMAIN` + optional `GATEWAY_TTY_DOMAIN`).
- Keep every `GATEWAY_TRAEFIK_*` knob as an explicit override for bring-your-own-Traefik topologies.

Ships in `v2026.05-rc9`.

PaulDotterer:
## Summary

Large consolidation PR addressing findings from the rc10 audit sweep. Breaking changes allowed per scope decision. Organized by theme:

### Security-critical fixes
- **Close `CONTROL_PASSWORD_AUTH_ENABLED` bypass.** The global flag previously only gated the SSO handler's \`ListAuthMethods\` response (a UI hint); the Login RPC itself never checked it, so direct POST-based password login succeeded against any account with \`has_password=true\` even in SSO-only deployments. Now enforced in \`AuthHandler.Login\` before the DB lookup, burning the dummy bcrypt cycle on the disabled path so timing doesn't leak whether password login is enabled. Regression test: \`TestLogin_GlobalPasswordAuthDisabled\`.
- **Terminal tokens are now single-use.** \`TokenStore.Validate\` uses a new atomic \`GetAndDelete\` backend primitive (Valkey GETDEL, in-memory CAS on the fake). A leaked bearer can no longer replay a second WebSocket within the TTL. Forgery attempts restore the unused entry with remaining TTL so a legitimate client isn't locked out by a guess. Regression tests: \`TestTokenStore_Validate_IsSingleUse\`, \`TestTokenStore_Validate_MismatchPreservesSession\`.

### Fail-open audit (gateway startup)
Six \`os.Exit(1)\` paths converted to \`Warn\`+feature-disabled so a malformed optional knob no longer takes agent mTLS down:
- Malformed \`PublicTerminalURLTemplate\` → terminal disabled
- \`registry.RegisterGateway\` failure → terminal disabled for replica
- \`routableIP()\` failure for \`GATEWAY_INTERNAL_URL\` → admin fan-out disabled
- \`routableIP()\` failure for Traefik backend → skip publish (don't poison Redis KV with empty backend)
- \`PublishTraefikRoute\` failure at startup → rely on refresh goroutine's retry cadence (the crash loop was exactly the rc8 failure mode)
- \`BootstrapHost\` without derivable assignedHost → disable middleware instead of crashing

Remaining \`os.Exit\` paths (TLS cert/key/CA flags, Valkey/Control URL missing, cert parse, \`tls.Config\`, \`http2.ConfigureServer\`) stay fatal — gateway cannot serve primary mTLS role without them.

### Data integrity
- **LUKS three-phase audit stream-ID coherence.** New sqlc query \`GetLuksRevocationStreamID\` looks up the \`luks_key\` event-stream ID minted at request time so the async Revoked/Failed event lands on the SAME stream as the Requested/Dispatched phases. Previous behavior generated a fresh ULID, splitting every revocation across two streams and breaking the projection stitch.
- **SecurityAlert projection.** Migration 010 installs \`security_alerts_projection\` + sidecar AFTER INSERT trigger. Events were persisted but nothing projected them; audit caught silent data loss on the SIEM path.

### Config coherence
- \`GATEWAY_TTY\` host now falls back through the full chain \`GATEWAY_TRAEFIK_TTY_HOST\` → \`GATEWAY_TTY_DOMAIN\` → \`GATEWAY_DOMAIN\`. \`.env.example\` always documented the third rung; \`config.go\` finally implements it (this was the root of the staging crash loop).
- Empty \`CONTROL_GATEWAY_URL\` is fatal at startup (was a warn); \`RegistrationHandler.Register\` has defense-in-depth refusal.
- \`deploy/compose.yml\` now passes \`GATEWAY_VALKEY_DB\` through (was dead config).

### Deploy hardening
- \`IMAGE_TAG\` now **required** via \`\${IMAGE_TAG:?…}\`, no more silent \`:latest\` upgrades on \`docker compose pull\`.
- \`redis-stack-server\` pinned to \`7.4.0-v0\` so upstream changes don't silently land.
- \`initdb.d/01-indexer-user.sh\` hard-fails on unset \`INDEXER_POSTGRES_PASSWORD\` instead of returning 0.

### Docs
- \`cmd/control/README.md\`: drop stale \`:8080\` from the \`CONTROL_GATEWAY_URL\` example (compose routes via \`:443\` with TCP passthrough).
- \`server/README.md\`: correct \`GATEWAY_WEB_LISTEN_ADDR\` description (cleartext HTTP behind Traefik, not TLS).

### Error-code parity guard
New unit test \`TestErrorCodeParityWithTSSDK\` walks \`errors.go\` and the sdk's \`ts/errors.ts\` via filesystem, asserts the two code sets match. Catches the drift the rc10 audit found (4 codes missing on TS side).

## Deferred to follow-ups (explicit non-goals for this PR)

The rc10 plan also called for:
- Revocation pub/sub (DeviceRevoked → close gateway stream) — significant new subsystem (Valkey pub/sub channel + gateway subscriber + stream close semantics). Worth its own design review and PR.
- URL-knob consolidation into \`GATEWAY_PUBLIC_URL\` — invasive config-shape change, deserves its own breaking-release PR.
- Legacy env alias fail-fast (\`TTY_DOMAIN\`, unprefixed \`VALKEY_*\`) — additive hardening that shouldn't block the security fixes in this PR.
- Integration-test harness covering the full fail-open matrix end-to-end.
- Multi-event append transactional wrapping (\`CreateUser\`, \`LinkOrCreate\`) — bigger refactor that touches several handlers.
- \`ON CONFLICT\` on rotation projections — projection replay is a theoretical concern not seen in practice.

Each of these has been captured on the rc10 audit notes and can land as focused follow-ups without the scope creep risk of piling them all into one PR.

## Audit claims verified and rejected

Publishing these explicitly so future devs don't chase ghosts:
- ❌ LPS plaintext password leak in event metadata: already encrypted at rest via \`h.encryptor.Encrypt(r.Password)\` in \`internal_handler.go:233\`.
- ❌ \`inbox_worker.go\` "appends both success AND failure events" for LUKS revocation: \`return AppendEvent(...)\` terminates the function; only one path runs.
- ❌ Expired cert acceptance on gateway mTLS: TLS layer + \`x509.Verify\` default \`time.Now()\` already gates this.
- ❌ \`CONTROL_ENCRYPTION_KEY_REQUIRED=False\` capital-F typo bypass: code uses literal \`== \"false\"\` compare; capital F falls into fail-closed branch.
- ❌ OSQuery bypass of event store: \`osquery_results\` is operational data, not a projection — intended design.
- ❌ Proto files out of sync with generated code: mtimes match.

Fixes #74.

## Test plan

- [x] \`go build ./...\` clean across control, gateway, indexer
- [x] \`go test ./internal/api/\` including \`TestLogin_GlobalPasswordAuthDisabled\`, \`TestErrorCodeParityWithTSSDK\`
- [x] \`go test ./internal/terminal/\` including single-use + mismatch-restore regression tests
- [ ] Full \`go test ./...\` including the long \`internal/api\` testcontainer run
- [ ] Staging redeploy: \`IMAGE_TAG=v2026.05-rc10 docker compose pull && docker compose up -d\`, agent reinstall, confirm mTLS handshake completes without cert-unknown
- [ ] Deliberately-broken \`.env\` dry-run: unset \`GATEWAY_TTY_DOMAIN\`, confirm gateway stays up and serves agent mTLS (fails-open)

## Summary by CodeRabbit

* **New Features**
  * Track, list, and acknowledge security alerts
  * Option to disable password authentication
  * Session tokens are single-use (consumed on successful validation)

* **Improvements**
  * Stricter gateway/config validation with redacted logging and startup checks
  * More resilient startup/registry behavior and safer Traefik publishing
  * Compose tightened (required image tag, pinned image); init scripts fail fast on missing secrets

* **Documentation**
  * Clarified gateway listener, host/port semantics, terminal token contract

* **Tests**
  * New parity and validation tests for errors, gateway URLs, config, and token semantics


PaulDotterer:
@Coderabbitai Full review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@codserabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

Issue #80, opened 2026-04-25
feat: bootstrap install.sh + guided setup.sh for fresh server deploys (rc11)


PaulDotterer:
## Problem

A fresh server deploy currently requires the operator to:

1. Find the latest release of `server/deploy/`, download or git clone it.
2. `cp .env.example .env`.
3. Edit `.env` by hand and fill in ~10 vars across multiple unrelated sections (database password, indexer-DB password with URL-safe constraint, Valkey password, JWT secret, admin email/password, control + gateway + TTY domains, ACME email, encryption key, IMAGE_TAG).
4. Run `setup.sh` to generate the internal CA + certs.
5. `docker compose up -d`.

Each step has a footgun (URL-unsafe characters in `INDEXER_POSTGRES_PASSWORD` produces an obscure parse error at indexer startup; `GATEWAY_TTY_DOMAIN` must differ from `GATEWAY_DOMAIN` if terminals are wanted; `CONTROL_ENCRYPTION_KEY` must be exactly 64 hex chars; etc.). Hard for new operators to get right on the first try, and the failures mostly surface long after `docker compose up` returns success.

## Design

Two complementary scripts in `server/deploy/`:

### Part A — `install.sh` (new bootstrap script)

Drops into a fresh Linux host with Docker already installed, brings up the full stack. **Does NOT install Docker** — checks for it and exits with a clear message if missing. Out of scope to handle the dozens of distro/version permutations of installing Docker; the operator sets that up first via their preferred channel (apt, dnf, official convenience script).

```bash
#!/usr/bin/env bash
# Power Manage server bootstrap.
#   curl -fsSL https://raw.githubusercontent.com/MANCHTOOLS/power-manage-server/main/deploy/install.sh | sudo bash
# or download + inspect first (recommended).
set -euo pipefail

INSTALL_DIR="${INSTALL_DIR:-/opt/power-manage}"
RELEASE_TAG="${RELEASE_TAG:-latest-rc}"   # matches the release-engineering image-tag contract

# Step 1 — preflight: Docker + compose plugin must be present and runnable.
require_cmd docker
require_compose_plugin   # `docker compose version`

# Step 2 — fetch the deploy/ tree from the chosen release tag.
#   (downloads a tarball from the GitHub releases endpoint;
#    no git dependency required on the host)
download_deploy_tree "$RELEASE_TAG" "$INSTALL_DIR"

# Step 3 — interactive env setup (delegates to setup.sh's new guided mode).
[...]

Issue #79, opened 2026-04-25
feat: distinct error code for 'gateway terminal URL not registered' (rc11)


PaulDotterer:
## Problem

`TerminalHandler.resolveGatewayURL` ([terminal_handler.go:274,279](../blob/main/internal/api/terminal_handler.go#L274)) returns `apiErrorCtx(ctx, ErrInternal, connect.CodeUnavailable, ...)` when the gateway hosting a device isn't registered (or the registry is unreachable). The gRPC code is `Unavailable` — appropriate — but the **string code** is `internal_error`, the generic catch-all.

Web's `errors.ts` keys off the string code, hits `internal_error`, and shows the generic toast "An unexpected error occurred. Please try again." The actual server-side message ("gateway hosting this device is no longer registered; retry shortly") never reaches the user.

Real impact (observed during rc10 deploy verification): a misconfiguration that produces this exact failure mode required ~30 min of `docker compose logs` + Valkey-key inspection to diagnose, when the toast itself could have said "the gateway hosting this device hasn't published its terminal URL — check `GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE`."

## Design

Introduce a dedicated error code for the registry-lookup-failed-but-gateway-known-to-control failure mode and surface it cleanly to the web client.

### Server side

```go
// internal/api/errors.go
const (
    // ... existing codes ...
    ErrGatewayNotRegistered = "gateway_not_registered"
)
```

In `terminal_handler.go:274` (the case where `LookupGatewayTerminalURL` returns `ErrNoGateway`):

```go
return "", apiErrorCtx(ctx, ErrGatewayNotRegistered, connect.CodeUnavailable,
    "gateway hosting this device is registered for the device-mapping but has not published its terminal URL — check GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE on the gateway")
```

The message is operator-facing (assumes the toast reader has shell access). Web translation can be friendlier.

`terminal_handler.go:279` (the registry-unavailable case) stays `ErrInternal` — that one IS a generic infra failure.

### SDK side

```typescript
// sdk/ts/errors.ts
export const ErrGatewayNotRegistered = 'gateway_not_registered';
```

### Web side

```json
// web/messages/en.json
"error_gateway_not_registered": "The gateway hosting this device hasn't published its terminal URL. Ask your administrator to verify that GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE is set on the gateway.",
```

```json
[...]

Issue #78, opened 2026-04-25
feat: gateway warns at startup when terminal config is partial (rc11)


PaulDotterer:
## Problem

When an operator partially configures the terminal feature in `.env`, the gateway silently skips terminal registration without logging anything. Concretely: `GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE` is the gate that controls whether the gateway publishes its terminal URL to the registry; if that one var is unset but the operator has already set `GATEWAY_WEB_LISTEN_ADDR` or `GATEWAY_TTY_DOMAIN` (clearly intending to enable terminals), the gateway boots happily, agents enroll, mTLS works — and the operator only discovers the misconfiguration when StartTerminal fails with a generic "An unexpected error occurred" toast hours later.

Surfaced during rc10 deploy on staging — took ~30 min of log spelunking and Valkey inspection to identify "you forgot to set the URL template." That diagnosis should be surfaceable from a single startup log line.

The current rc10 fail-open behavior is right for *malformed* templates (an empty-host URL is logged as a Warn). It's wrong for *empty* templates when other terminal env vars are set — that's a partial config, not a deliberate-feature-off, and partial config deserves a Warn.

## Design

Extend `Config.Validate()` (mirror the TTY/MTLS collision check pattern from rc10) to detect the partial-terminal-config combinations.

Three relevant env vars — `GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE`, `GATEWAY_WEB_LISTEN_ADDR`, `GATEWAY_TTY_DOMAIN` (or its fallback `GATEWAY_DOMAIN` is fine since that one is always set in practice). When any subset of these three is set but at least one is missing, emit a Warn (NOT a fatal — terminal feature stays optional) naming the missing piece(s):

```go
// in Config.Validate(), after the TTY/MTLS collision check
if c.WebListenAddr != "" && c.PublicTerminalURLTemplate == "" {
    return fmt.Errorf("warning-only: GATEWAY_WEB_LISTEN_ADDR is set "+
        "(=%q) but GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE is empty — "+
        "terminal sessions will not work because the gateway won't "+
        "publish its terminal URL to the registry; set the template "+
        "to enable, or unset GATEWAY_WEB_LISTEN_ADDR if you don't want terminals",
        c.WebListenAddr)
}
```

[...]

PR #83, opened 2026-04-25
feat: rc11 consolidation — derived system actions, terminal-error split, install.sh, guided setup

### LINKED ISSUE #77: feat: system actions as a derived projection — event listener + periodic reconciler safety net
## Problem

`SyncUserSystemActions` — which provisions `pm-tty-<username>`, the provisioning USER action, and the SSH access action on a user's assigned devices — is only called from 8 sites in `internal/api/user_handler.go`. Every other mutation that changes a user's effective permission set or system-action-relevant metadata appends its event and returns without syncing. The full drift surface:

| Event | Current handler | Calls sync? |
|-------|-----------------|-------------|
| `UserRoleAssigned` (role_handler standalone) | `role_handler.go:315` | ❌ |
| `UserRoleRevoked` | `role_handler.go:382` | ❌ |
| `RoleUpdated` (permissions changed) | `role_handler.go:185` | ❌ (every holder) |
| `RoleDeleted` | `role_handler.go:249` | ❌ (every holder) |
| `UserGroupMemberAdded` | `user_group_handler.go:343` | ❌ |
| `UserGroupMemberRemoved` | `user_group_handler.go:406` | ❌ |
| `UserGroupRoleAssigned` | `user_group_handler.go:485` | ❌ (every member) |
| `UserGroupRoleRevoked` | `user_group_handler.go:531` | ❌ (every member) |
| `UserGroupDeleted` | `user_group_handler.go:263` | ❌ (every member) |
| `UserGroupQueryUpdated` (dynamic) | `user_group_handler.go:606` | ❌ (affected members) |
| `ServerSettingUpdated` provisioning/SSH toggle | `settings_handler.go:51` | partial (fire-and-forget goroutine) |
| SSO auto-create + role assign | `idp/linker.go:200,225` | ❌ |
| SSO group map add/remove | `idp/linker.go:330,345` | ❌ |
| SCIM create + role assign | `scim/users.go:291,336` | ❌ |
| SCIM disable/enable/profile update | `scim/users.go:512,525,544,683,692,764,1065,1074` | ❌ |
| SCIM delete user | `scim/users.go:852` | ❌ (no `CleanupDeletedUserActions`) |

### Observed failure mode

A user granted the Admin role via `AssignRoleToUser` (rather than at creation time) holds the `StartTerminal` permission but never gets `pm-tty-<username>` deployed to their assigned devices. `StartTerminal` RPC succeeds, WebSocket opens, agent tries to switch to a TTY user that doesn't exist, session fails.

Nothing self-heals this except a control-server restart (which runs `SyncAllUsersSystemActions` at startup, [cmd/control/main.go:350](../blob/main/cmd/control/main.go#L350)) or an operator toggle of global provisioning/SSH settings.

## Design principle — system actions are a derived projection

Treat `pm-tty-*`, the USER provision action, and the SSH access action as a **pure projection** of source-of-truth state (user attributes + direct role assignments + user-group memberships + role permissions + global server settings). Handlers mutate source state only; `SystemActionManager` is the single authority that computes and reconciles desired-vs-actual. No handler ever calls the manager.

Two layers together guarantee convergence:

1. **Event-driven listener** on `Store.AppendEvent` for sub-second convergence. Classifies incoming events into affected user IDs and enqueues per-user sync. Replaces all scattered `SyncUserSystemActions` calls.
2. **Periodic full reconciler** on a short interval as the durability safety net. Covers the gap where the listener fires post-commit but the process dies before the sync runs, plus any event type that gets added to the source schema without being added to the classifier.

This mirrors the project's existing drift-prevention template in [`internal/search/index.go`](../blob/main/internal/search/index.go) (event-driven enqueue + periodic `Rebuild`), applied cleanly to system actions with the derived-model discipline that the search indexer doesn't fully adopt.

### 1. Store-side event listener hook

`Store.AppendEvent` gains a simple listener registration:

```go
// internal/store/listeners.go (new)
type EventListener func(ctx context.Context, event Event)

func (s *Store) RegisterEventListener(fn EventListener) {
    s.listeners = append(s.listeners, fn)
}
```

`AppendEvent` invokes registered listeners **post-commit, fire-and-forget in a goroutine**. A slow listener must never stall the write; a crashing listener must never roll back the event.

### 2. Event→affected-users classifier

Centralized in `internal/api/system_actions_listener.go`. One switch statement, one place to maintain the event→user-id map:

```go
type SyncOp int
const (
    SyncOpNone SyncOp = iota
    SyncOpSyncUser       // SyncUserSystemActions(userID)
    SyncOpCleanupUser    // CleanupDeletedUserActions(userProjection)
    SyncOpSyncAll        // SyncAllUsersSystemActions — for fan-out events
)

// AffectedFromEvent returns the op + user IDs this event triggers.
// Fan-out events (RoleUpdated, GroupDeleted, ServerSettingUpdated) return
// SyncOpSyncAll rather than materialising a potentially-large user list
// in the classifier path — the manager is already efficient for the full
// sweep and one sweep is cheaper than 10k per-user enqueues on a big fleet.
func AffectedFromEvent(e store.Event) (SyncOp, []string) {
    switch e.EventType {
    case "UserCreated", "UserRoleAssigned", "UserRoleRevoked",
         "UserGroupMemberAdded", "UserGroupMemberRemoved",
         "UserDisabled", "UserEnabled",
         "UserLinuxUsernameChanged",
         "UserProvisioningSettingsUpdated", "UserSshSettingsUpdated",
         "UserProfileUpdated", "UserSshKeyAdded", "UserSshKeyRemoved":
        return SyncOpSyncUser, []string{userIDFromEvent(e)}

    case "UserDeleted":
        return SyncOpCleanupUser, []string{e.StreamID}

    case "RoleUpdated", "RoleDeleted",
         "UserGroupRoleAssigned", "UserGroupRoleRevoked",
         "UserGroupDeleted", "UserGroupQueryUpdated",
         "ServerSettingUpdated":
        return SyncOpSyncAll, nil

    default:
        return SyncOpNone, nil
    }
}
```

### 3. Listener wiring

Registered once at service boot in `cmd/control/main.go`:

```go
st.RegisterEventListener(func(ctx context.Context, e store.Event) {
    op, userIDs := systemactions.AffectedFromEvent(e)
    switch op {
    case systemactions.SyncOpSyncUser:
        for _, uid := range userIDs {
            if err := svc.SystemActions().SyncUserSystemActions(ctx, uid); err != nil {
                logger.Error("listener: sync user failed", "user_id", uid, "event", e.EventType, "error", err)
            }
        }
    case systemactions.SyncOpCleanupUser:
        for _, uid := range userIDs {
            user, err := st.Queries().GetUserByID(ctx, uid)
            if err != nil {
                logger.Error("listener: load user for cleanup failed", "user_id", uid, "error", err)
                continue
            }
            if err := svc.SystemActions().CleanupDeletedUserActions(ctx, user); err != nil {
                logger.Error("listener: cleanup user failed", "user_id", uid, "error", err)
            }
        }
    case systemactions.SyncOpSyncAll:
        if err := svc.SystemActions().SyncAllUsersSystemActions(ctx); err != nil {
            logger.Error("listener: sync all users failed", "event", e.EventType, "error", err)
        }
    }
})
```

### 4. Periodic reconciler (safety net)

Listeners are post-commit and not durable — a process crash between the event commit and the listener running loses the trigger for that event. The periodic sweep catches any such gap plus any future event type that gets added to the schema without being added to `AffectedFromEvent`.

```go
// internal/api/system_actions.go

// StartReconciliation launches a background goroutine that periodically
// runs SyncAllUsersSystemActions as the durability safety net for the
// event-driven listener (see internal/api/system_actions_listener.go).
//
// Guards:
//   - atomic flag prevents a slow sweep from starting a second one
//     behind it (large fleet + transient DB slowness scenarios)
//   - per-sweep timeout prevents a hung query from blocking forever
//     and piling up missed ticks
//   - SyncAllUsersSystemActions is already best-effort per user, so
//     one bad user cannot abort the sweep
func (m *SystemActionManager) StartReconciliation(ctx context.Context, interval, sweepTimeout time.Duration) {
    go func() {
        ticker := time.NewTicker(interval)
        defer ticker.Stop()

        var running atomic.Bool

        for {
            select {
            case <-ctx.Done():
                return
            case <-ticker.C:
                if !running.CompareAndSwap(false, true) {
                    m.logger.Warn("skipping system-action reconciliation tick — previous sweep still running")
                    continue
                }
                sweepCtx, cancel := context.WithTimeout(ctx, sweepTimeout)
                if err := m.SyncAllUsersSystemActions(sweepCtx); err != nil {
                    m.logger.Error("periodic system-action reconciliation failed", "error", err)
                }
                cancel()
                running.Store(false)
            }
        }
    }()
}
```

Wired in `cmd/control/main.go` right after the startup sync:

```go
if cfg.SystemActionReconcileInterval > 0 {
    svc.SystemActions().StartReconciliation(ctx,
        cfg.SystemActionReconcileInterval,
        cfg.SystemActionReconcileTimeout,
    )
    logger.Info("system-action reconciliation started",
        "interval", cfg.SystemActionReconcileInterval,
        "sweep_timeout", cfg.SystemActionReconcileTimeout)
}
```

### 5. Calm logs

`SyncAllUsersSystemActions` currently emits an `Info` on every successful sweep ([system_actions.go:54](../blob/main/internal/api/system_actions.go#L54)). Demote to `Debug` so periodic ticks don't spam. The startup caller in `cmd/control/main.go:350` gets its own `Info` line so the one-shot startup sweep remains visible.

### 6. Handler-side deletions

All existing scattered `SyncUserSystemActions` call sites in `internal/api/user_handler.go` (8 of them: CreateUser, disable/enable, profile update, provisioning toggle, linux username change, SSH key add/remove, SSH settings update) are **removed**. They become redundant once the listener covers every permission-shaping event — keeping them risks double-runs on every mutation.

Explicitly NOT adding "fast-path" sync calls to role/group/SCIM handlers. The derived-model invariant is `handlers mutate source state only; the manager owns system actions`. Fast-path hooks would violate that invariant and reintroduce the sibling-sweep problem this design eliminates.

## Config

Two new env vars, documented in `deploy/.env.example`:

| Var | Default | Effect |
|-----|---------|--------|
| `CONTROL_SYSTEM_ACTION_RECONCILE_INTERVAL` | `1m` | Period between full sweeps. `0` disables periodic reconciliation. |
| `CONTROL_SYSTEM_ACTION_RECONCILE_TIMEOUT` | `5m` | Per-sweep context deadline. |

1-minute default because the sweep is read-heavy, idempotent, short-circuits on no-op users, and serves primarily as the safety net. Expected latency for the common case is listener-time (sub-second); the 1-minute interval is how long we'd wait if the listener dropped the signal.

## Files touched

- `internal/store/listeners.go` (new): `EventListener` type + `RegisterEventListener` method.
- `internal/store/store.go`: invoke listeners post-commit from `AppendEvent`.
- `internal/api/system_actions.go`: add `StartReconciliation`; demote line 54 success log to `Debug`.
- `internal/api/system_actions_listener.go` (new): `SyncOp` enum + `AffectedFromEvent` classifier + `userIDFromEvent` helper.
- `cmd/control/main.go`: parse two new env vars; register the listener; call `StartReconciliation`; add an `Info` log line for the one-shot startup sweep.
- `internal/api/user_handler.go`: **remove** all 8 `SyncUserSystemActions` call sites.
- `deploy/.env.example`: document `CONTROL_SYSTEM_ACTION_RECONCILE_INTERVAL` and `CONTROL_SYSTEM_ACTION_RECONCILE_TIMEOUT`.

Note: `internal/api/role_handler.go`, `internal/api/user_group_handler.go`, `internal/api/settings_handler.go`, `internal/idp/linker.go`, and `internal/scim/users.go` are **NOT touched**. The listener covers every permission-shaping event they emit; adding sync calls there would violate the derived-model invariant.

## Test plan

- **Unit** (`internal/api`): `AffectedFromEvent` — one case per supported event type, confirm correct `SyncOp` + user-id set. One case for each unsupported event type, confirm `SyncOpNone`. One case per fan-out event, confirm `SyncOpSyncAll` with empty user list.
- **Unit** (`internal/api`): `StartReconciliation` fires `SyncAllUsersSystemActions` on the supplied interval; overlapping-tick guard skips with a Warn when the prior sweep hasn't finished; per-sweep timeout cancels a stuck invocation without leaking the goroutine; `ctx.Done()` stops the goroutine cleanly.
- **Unit** (`internal/store`): listener registration; post-commit invocation; a panicking listener does not affect other listeners or the append return value.
- **Integration (testcontainers)**: for each of the 15 event types in the problem-statement table, emit the event via the corresponding handler, assert `SyncUserSystemActions` ran for the affected user within N ms (N ~= 100ms for listener-path events) and produced the expected action state.
- **Integration (testcontainers)**: fan-out event — e.g. `RoleUpdated` adding `StartTerminal` — assert every holder gets the TTY action within one listener cycle.
- **Integration (testcontainers)**: SCIM delete user — assert `CleanupDeletedUserActions` ran, system TTY action for that user is removed, no orphan entries remain in the projection.
- **Integration**: start fresh, `AssignRoleToUser` + admin role (the failure mode from this issue), assert TTY action is created and dispatched within one listener cycle without any other mutation.
- **Integration**: simulate listener drop — register a listener that ignores the target event type, advance the clock past the reconcile interval, confirm drifted users converge via the periodic sweep.
- **Integration**: with `CONTROL_SYSTEM_ACTION_RECONCILE_INTERVAL=0`, confirm the periodic goroutine is not started but the listener still works.
- **Regression**: full `go test ./internal/api/...` including existing handler tests — removing the 8 `SyncUserSystemActions` calls must not regress `TestCreateUser_*`, disable/enable, profile update, or SSH key tests.

## Acceptance criteria

1. **Derived-model invariant**: no file outside `internal/api/system_actions*.go` and `cmd/control/main.go` imports `SystemActionManager` or calls `SyncUserSystemActions` / `CleanupDeletedUserActions`. A lint rule or a `TestSystemActionsImportInvariant` test enforces this.
2. Granting or revoking a role via `AssignRoleToUser` / `UnassignRoleFromUser` provisions or cleans up the TTY system action within the listener cycle (sub-second).
3. Adding a user to a user-group that holds `StartTerminal` provisions the TTY action within the listener cycle without any other mutation.
4. SCIM-provisioned user with a role mapping gets the TTY action within the listener cycle of the `UserRoleAssigned` event landing.
5. `RoleUpdated` adding or removing `StartTerminal` from a role reflects on all holders within the listener cycle (fan-out path).
6. SCIM user deletion runs `CleanupDeletedUserActions` — system actions for the deleted user are cleaned up, no orphans remain.
7. A control server running for 24h with a drift-prone workload has no unreconciled system-action state, verifiable by diffing the projection against a fresh `SyncAllUsersSystemActions` output.
8. Periodic reconciliation logs: nothing on success, `Warn` on skipped tick (overlap), `Error` on failure. Startup sync retains its `Info` line. Listener logs errors only.
9. Setting the interval to `0` disables the periodic goroutine entirely without regressing the listener path.
10. All 8 `SyncUserSystemActions` call sites in `internal/api/user_handler.go` are deleted; handler tests still pass.

## Out of scope

- **Per-user dedup / debounce queue.** A SCIM batch import firing 20 events for one user within 500ms enqueues 20 sync calls. `SyncUserSystemActions` is idempotent, so the cost is redundant DB reads, not correctness. Worth adding later if benchmarks show it matters; listed here so it's not forgotten. ~60 lines in a follow-up.
- **Redesigning `SyncAllUsersSystemActions` for per-user concurrency.** Currently serial; benchmark before optimising.
- **Applying the same derived-model pattern to the search indexer's scattered `EnqueueReindex` call sites.** Same class of problem, same solution shape, but a larger surface — separate issue.
- **Asynq task fan-out for the listener.** Post-commit goroutines are sufficient and avoid a new queue dependency. Revisit only if listener latency or durability becomes an operational concern.

PaulDotterer:
## Summary

Bundles the four rc11 follow-ups so CodeRabbit gets one review per repo instead of four small ones.

### #77 — Derived-projection system actions

Replaces the 8 scattered `SyncUserSystemActions` calls in user/SCIM handlers with a single store-side event listener. The listener classifies each persisted event into a `SyncOp` (per-user / per-cleanup / fan-out-all) via the new `AffectedFromEvent` function and dispatches asynchronously. A periodic reconciler (`StartReconciliation`, default 1m) is the safety net for events the classifier has not yet learned about, matching the existing search-indexer derived-projection pattern.

The listener spawns a goroutine per dispatch (so AppendEvent's post-commit path is never blocked) and bounds each goroutine with `CONTROL_SYSTEM_ACTION_RECONCILE_TIMEOUT` (default 5m) so a wedged DB or signer cannot leak goroutines indefinitely. `UserDeleted` is deliberately NOT classified — `CleanupDeletedUserActions` needs the user projection loaded BEFORE the event is applied, so it stays in the handler that emits the event (with SCIM now using the same pattern via the new `SystemActionsCleaner` interface).

### #78 — Partial-terminal-config warn

`config.Validate()` now returns `(warnings []string, err error)` and warns when only a subset of the three terminal vars is set, since silently treating that as \"terminals disabled\" caused operator confusion. Added `TTYDomainExplicitlySet` so the auto-derived URL template can be distinguished from an operator-set one.

### #79 — `gateway_not_registered` error code

Splits the previous \"device not connected\" overloading. When `LookupGatewayDial` returns `ErrNoGateway`, the handler now probes `LookupGatewayInternalURL`:
- both fail → gateway is gone (`device_not_connected`, transient)
- internal URL exists, dial doesn't → gateway is alive but never published its terminal URL (`gateway_not_registered`, persistent misconfig — surfaces a hint to set `GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE`)

Paired with `manchtools/power-manage-sdk` and `manchtools/power-manage-web` PRs for the const + i18n key.

### #80 — install.sh + guided setup.sh

New `deploy/install.sh` bootstrap installer. Refactored `deploy/setup.sh` with an interactive guided env loop (`prompt_secret`, `prompt_string`, `prompt_yes_no`), a `clear_env_var` helper, and a No-branch on the terminal prompt that explicitly clears `GATEWAY_TTY_DOMAIN`, `GATEWAY_PUBLIC_TERMINAL_URL_TEMPLATE`, and `GATEWAY_WEB_LISTEN_ADDR` on rerun (the previous cut silently no-op'd, leaving stale values from the first run). `--no-prompt` and non-TTY stdin preserve the CI path.

## Review-round fixes already in this PR

- Async listener (review round 1): the first cut invoked `SyncAllUsersSystemActions` synchronously inside the AppendEvent post-commit hook, which would have made small admin RPCs O(all-users). Now goroutine-dispatched.
- Goroutine bounds (review round 2): `context.Background()` + no timeout would have leaked one goroutine per event under DB/signer wedge. Now bounded by `CONTROL_SYSTEM_ACTION_RECONCILE_TIMEOUT`.
- Terminal error split (review round 1): see #79 above. Earlier version returned the same `device_not_connected` code for both gateway-gone and gateway-alive-no-terminal-URL, hiding a config bug behind a transient-retry message.
- setup.sh disable-clears (review round 1): see #80 above.

## Test plan

- [x] `go build ./...` clean across server / agent / sdk
- [x] `go test ./internal/config/` — ok
- [x] `go test -run AffectedFromEvent ./internal/api/` — 15 cases pass (every classifier branch + defensive missing-user_id)
- [x] `go test -run TestStartTerminal_GatewayNotRegistered ./internal/api/` — 2 cases pass (persistent misconfig vs transient loss)
- [x] `bash deploy/setup_test.sh` — 6 cases pass (write_env_var add/update/preserve, clear_env_var existing/missing, disable-terminals-clears-three-vars)
- [ ] CodeRabbit review

Closes #77, #78, #79, #80.

## Summary by CodeRabbit

* **New Features**
  * Optional automatic periodic system-action reconciliation with configurable interval/timeout; permission-change events now trigger async sync.
  * New installer with guided and non-interactive setup, plus a quickstart bootstrap flow.

* **Bug Fixes**
  * SCIM user deletion performs best-effort cleanup of related system actions.
  * Terminal/gateway errors now return clearer, actionable hints.

* **Documentation**
  * Added comprehensive quickstart and install guidance.

* **Tests**
  * New tests covering event-to-sync classification and terminal gateway error cases.

### DIFF
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -375,8 +375,15 @@ func main() {
 
 		// (2) Listener — registered post-commit on the store. Logged
 		// errors are swallowed; the periodic reconciler is the
-		// durability safety net.
-		st.RegisterEventListener(api.SystemActionListener(svc.SystemActions(), st, logger.With("component", "system_action_listener")))
+		// durability safety net. Reuse the same per-sweep timeout
+		// as the reconciler so a wedged DB / signer can't leak a
+		// goroutine indefinitely (#77 review round 2).
+		st.RegisterEventListener(api.SystemActionListener(
+			svc.SystemActions(),
+			st,
+			logger.With("component", "system_action_listener"),
+			cfg.SystemActionReconcileTimeout,
+		))
 
 		// (3) Periodic reconciler — interval and per-sweep timeout
 		// from config (defaults set in parseFlags).
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -84,6 +84,13 @@ type Config struct {
 	ValkeyPassword string
 	ValkeyDB       int
 
+	// rc11 #77: derived-projection reconciler for system actions.
+	// Interval is the period between full SyncAllUsersSystemActions
+	// sweeps (safety net for the post-commit listener); 0 disables
+	// the periodic goroutine entirely. Timeout is the per-sweep
+	// context deadline so a hung query can't pile up missed ticks.
+	SystemActionReconcileInterval time.Duration
+	SystemActionReconcileTimeout  time.Duration
 }
 
 func main() {
@@ -345,11 +352,40 @@ func main() {
 		logger.Error("failed to reconcile system roles", "error", err)
 	}
 
-	// Sync system actions for all users at startup (idempotent)
+	// rc11 #77: derived-projection wiring for system actions.
+	//
+	// 1) One-shot startup sweep — guarantees idempotent convergence
+	//    on every boot, deploy, or upgrade. Logged at Info because it
+	//    runs once.
+	// 2) Post-commit event listener — fires SyncUserSystemActions
+	//    (or SyncAllUsersSystemActions for fan-out events) on every
+	//    permission-shaping event, so handler tests don't need to
+	//    know about system actions.
+	// 3) Periodic reconciler — durability safety net for the listener,
+	//    catches any event whose effect on system actions the
+	//    listener doesn't yet know about. Default 1m.
 	if svc.SystemActions() != nil {
+		// (1) Startup sweep — keeps the existing Info line so
+		// operators see the one-shot convergence in boot logs.
 		if err := svc.SystemActions().SyncAllUsersSystemActions(ctx); err != nil {
 			logger.Error("failed to sync system actions at startup", "error", err)
+		} else {
+			logger.Info("system actions synced for all users (startup)")
 		}
+
+		// (2) Listener — registered post-commit on the store. Logged
+		// errors are swallowed; the periodic reconciler is the
+		// durability safety net.
+		st.RegisterEventListener(api.SystemActionListener(svc.SystemActions(), st, logger.With("component", "system_action_listener")))
+
+		// (3) Periodic reconciler — interval and per-sweep timeout
+		// from config (defaults set in parseFlags).
+		svc.SystemActions().StartReconciliation(ctx,
+			cfg.SystemActionReconcileInterval,
+			cfg.SystemActionReconcileTimeout)
+		logger.Info("system-action reconciliation started",
+			"interval", cfg.SystemActionReconcileInterval,
+			"sweep_timeout", cfg.SystemActionReconcileTimeout)
 	}
 	// Configure trusted proxies for X-Forwarded-For header validation
 	if len(cfg.TrustedProxies) > 0 {
@@ -584,8 +620,10 @@ func main() {
 	path, handler := pmv1connect.NewControlServiceHandler(svc, interceptors)
 	mux.Handle(path, handler)
 
-	// Mount SCIM v2 handler
-	scimHandler := scim.NewHandler(st, logger)
+	// Mount SCIM v2 handler. Passes svc.SystemActions() so the SCIM
+	// delete path can clean up pm-tty-* / USER actions when the
+	// last identity link is removed (rc11 #77).
+	scimHandler := scim.NewHandler(st, logger, svc.SystemActions())
 	mux.Handle("/scim/v2/", scimHandler)
 
 	// Wrap with CORS and security headers middleware
@@ -732,6 +770,11 @@ func parseFlags() *Config {
 	flag.StringVar(&cfg.GatewayURL, "gateway-url", "", "Gateway URL returned to agents during registration")
 	flag.StringVar(&cfg.TerminalGatewayURL, "terminal-gateway-url", "", "Public WebSocket URL of the gateway terminal endpoint (e.g. wss://gw.example.com/terminal). When empty, ControlService.StartTerminal returns CodeUnavailable.")
 	flag.DurationVar(&cfg.DynamicGroupEvalInterval, "dynamic-group-eval-interval", time.Hour, "Interval for evaluating dynamic groups (min 30m, max 8h, 0 to disable)")
+	// rc11 #77 — system-action reconciliation defaults: 1m interval keeps drift bounded for an
+	// operator-visible UX path (role grant → terminal works), 5m sweep timeout is plenty for
+	// even a 10k-user fleet because the sync is read-heavy and short-circuits on no-op users.
+	flag.DurationVar(&cfg.SystemActionReconcileInterval, "system-action-reconcile-interval", time.Minute, "Period between full SyncAllUsersSystemActions sweeps; 0 disables periodic reconciliation")
+	flag.DurationVar(&cfg.SystemActionReconcileTimeout, "system-action-reconcile-timeout", 5*time.Minute, "Per-sweep context deadline for the periodic reconciler")
 	flag.StringVar(&cfg.CATrustBundlePath, "ca-trust-bundle", "", "PEM file with trusted CA certificates for verification (supports CA rotation)")
 	flag.BoolVar(&cfg.TLSEnabled, "tls", false, "Enable TLS on public listener")
 	flag.StringVar(&cfg.TLSCert, "tls-cert", "", "TLS certificate for public listener")
@@ -763,6 +806,8 @@ func parseFlags() *Config {
 	envString(&cfg.TerminalGatewayURL, "CONTROL_TERMINAL_GATEWAY_URL")
 	envCSV(&cfg.CORSOrigins, "CONTROL_CORS_ORIGINS")
 	envDuration(&cfg.DynamicGroupEvalInterval, "CONTROL_DYNAMIC_GROUP_EVAL_INTERVAL")
+	envDuration(&cfg.SystemActionReconcileInterval, "CONTROL_SYSTEM_ACTION_RECONCILE_INTERVAL")
+	envDuration(&cfg.SystemActionReconcileTimeout, "CONTROL_SYSTEM_ACTION_RECONCILE_TIMEOUT")
 
 	// SSO / Identity Provider configuration
 	cfg.PasswordAuthEnabled = true // default enabled
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -841,6 +841,19 @@ func parseFlags() *Config {
 		}
 	}
 
+	// Clamp system-action reconcile flags. Mirrors the DynamicGroup
+	// pattern above. A 0 sweep timeout would make
+	// context.WithTimeout return an already-cancelled context every
+	// tick, silently breaking the durability safety net; a negative
+	// interval would panic time.NewTicker. Round-3 review of rc11
+	// #77 caught the timeout footgun specifically.
+	if cfg.SystemActionReconcileInterval < 0 {
+		cfg.SystemActionReconcileInterval = 0 // treat as disabled, matching StartReconciliation
+	}
+	if cfg.SystemActionReconcileTimeout <= 0 {
+		cfg.SystemActionReconcileTimeout = 5 * time.Minute
+	}
+
 	if cfg.JWTSecret == "" {
 		fmt.Fprintln(os.Stderr, "FATAL: CONTROL_JWT_SECRET (or -jwt-secret) is required")
 		os.Exit(1)
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -512,24 +512,41 @@ func main() {
 
 		// Index audit events on insertion — the hook fires after every AppendEvent
 		// and enqueues the persisted row directly (no DB lookup in the search worker).
-		// Registered via RegisterEventListener (was st.OnEventAppended) so it's
-		// guarded by the same mutex as every other listener and the registration
-		// order is explicit: appears AFTER the system-action listener at line 381,
-		// so events that fire between the two are still picked up by the
-		// AppendEvent path because both registrations complete before the RPC
-		// servers start serving below.
+		// Registered via RegisterEventListener so it shares the listener-slice
+		// mutex + panic-recovery wrapper with every other consumer.
+		//
+		// The EnqueueReindex call itself is dispatched in a goroutine so a slow
+		// or unreachable Valkey cannot stall AppendEvent — fireListeners
+		// dispatches synchronously, so a blocking listener body would extend
+		// every state-changing RPC's tail latency by the Valkey RTT. The work
+		// is best-effort (already only logs Warn on failure), so detaching is
+		// safe; the goroutine has its own recover so a panic inside the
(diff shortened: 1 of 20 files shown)


PaulDotterer:
@coderabbitai full review

PaulDotterer:
@coderabbitai full review

PaulDotterer:
@coderabbitai full review

PR #87, opened 2026-04-26
fix: terminal access for unassigned devices via permission-derived TTY resolution

PaulDotterer:
### 1. `StartTerminal` ignored permission scope

[`terminal_handler.go:110`](https://github.com/manchtools/power-manage-server/blob/main/internal/api/terminal_handler.go#L110) hardcoded `filterUserID := userCtx.ID`, so even an admin with the unrestricted `StartTerminal` permission saw \"device not found\" on any device they hadn't directly assigned to themselves. Every other handler that reads from `devices_projection` uses the canonical `userFilterID(ctx, \"StartTerminal\")` pattern (cf. `device_handler.go:182`); this PR fixes the inconsistency.

### 2. TTY accounts only landed on user-assigned devices

`syncTtyUserAction` shipped the `pm-tty-<username>` action through an ordinary `AssignmentCreated` event targeting the user themselves ([`system_actions.go:419`](https://github.com/manchtools/power-manage-server/blob/main/internal/api/system_actions.go#L419)). Delivery therefore depended on the user holding device assignments — exactly backwards for admins who manage the fleet without being individually assigned to each device. Bulk-enrolled (unassigned) devices never received the TTY account, so even with the handler fix above, `StartTerminal` minted a token but the agent had nobody to log in as.

## Approach — resolve TTY actions from permissions, not assignments

New sqlc query `ListSystemTtyActionsForPermissionHolders` joins `users_projection.system_tty_action_id` against role/group permissions, returning every TTY action whose owner holds `StartTerminal` (directly or via group role). `DISTINCT ON (a.id)` collapses overlapping grants. Filtering hides soft-deleted users, actions, roles, and groups so stale projections can't leak.

`ResolveActionsForDevice` adds this as a fourth layer, merged after the device- and user-layer assignment sources:

- Deduped against everything already in the result.
- **Ignores device-layer `EXCLUDED`** — terminal access is the system's escape hatch and must not be turn-off-able from an operator-side exclusion. User deletion drives cleanup through the existing path (the system action itself is removed, agents drop the account on next sync), which is the safety property we rely on instead.

`syncTtyUserAction` no longer emits `AssignmentCreated` — delivery is purely permission-derived. Existing TTY actions keep their stale assignments harmlessly; the new resolver source covers them.

## Tests

- `TestStartTerminal_AdminUnassigned` — handler fix; admin can start a session on an unassigned device.
- `TestResolveActions_TTYPermissionSource_DirectRole` — TTY ships when `StartTerminal` is directly granted.
- `TestResolveActions_TTYPermissionSource_ViaUserGroup` — same, via group → role → permission chain.
- `TestResolveActions_TTYPermissionSource_NoPermissionExcluded` — linked-but-unprivileged user does NOT ship a TTY action (the cleanup-safety property).
- `TestResolveActions_TTYPermissionSource_DedupesAcrossRoles` — direct + group grants of the same permission collapse to one row.

All pass against testcontainers Postgres locally; existing resolution / terminal / registration suites unchanged.

## Audit/history note

New TTY actions stop emitting `AssignmentCreated`. That's an intentional semantic change in line with the design (delivery is permission-derived, not assignment-derived); the link to the user via `system_tty_action_id` and the action-creation event itself remain auditable.

## Summary by CodeRabbit

* **New Features**
  * Permission-derived TTY actions added so authorized users can start terminals on unassigned devices.

* **Bug Fixes**
  * TTY linking no longer emits assignment events.
  * Device-not-found messaging simplified.
  * Device-level exclusions no longer block permission-derived TTY actions.
  * StartTerminal authorization now respects unified user filtering.

* **Tests**
  * Added coverage for permission-sourced TTY resolution, deduplication, failure propagation, and terminal starts for unassigned devices.

### DIFF
--- a/internal/api/system_actions.go
+++ b/internal/api/system_actions.go
@@ -410,14 +410,17 @@ func (m *SystemActionManager) syncTtyUserAction(ctx context.Context, user db.Use
 	actionName := "system:tty-user:" + user.ID
 
 	if user.SystemTtyActionID == "" {
-		// Create new system action
+		// Create new system action. We deliberately do NOT emit an
+		// AssignmentCreated event for the TTY action — delivery is
+		// driven by the permission-derived path in resolution.go,
+		// which materializes the pm-tty-<username> account on every
+		// device whenever the user holds the StartTerminal permission.
+		// Coupling delivery to a per-user assignment was the original
+		// bug: admins manage the fleet without being assigned to any
+		// individual device, so their TTY accounts never landed.
 		actionID, err := m.createSystemAction(ctx, actionName, int32(pm.ActionType_ACTION_TYPE_USER), int32(pm.DesiredState_DESIRED_STATE_PRESENT), paramsJSON)
 		if err != nil {
 			return fmt.Errorf("create tty user action: %w", err)
-		}
-
-		if err := m.assignActionToUser(ctx, actionID, user.ID); err != nil {
-			return fmt.Errorf("assign tty user action: %w", err)
 		}
 
 		if err := m.linkSystemAction(ctx, user.ID, "system_tty_action_id", actionID); err != nil {
--- a/internal/api/system_actions.go
+++ b/internal/api/system_actions.go
@@ -15,10 +15,24 @@ import (
 )
 
 // SystemActionManager creates and maintains system-managed actions
-// (user provisioning and SSH access) for PM users. These actions are
-// stored in actions_projection and assigned to users via
-// assignments_projection, so the resolution engine picks them up
-// automatically for devices with assigned users.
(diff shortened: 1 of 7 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbit review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbit review

PR #91, opened 2026-05-03
feat: container-wins resolution and grouped sync wire

PaulDotterer:
## Summary

Server half of the agent-side ordering rework for [manchtools/power-manage-agent#45](https://github.com/manchtools/power-manage-agent/issues/45). Built on top of the SDK changes that landed in [manchtools/power-manage-sdk#40](https://github.com/manchtools/power-manage-sdk/pull/40).

### Schedule on containers

* `action_sets_projection` and `definitions_projection` gain a required `schedule JSONB` column (migration 012). Default `{"interval_hours": 8}` mirrors the existing per-action default cadence so any pre-existing rows behave the same as before once they're delivered through the new sync wire shape.
* New event types `ActionSetScheduleUpdated` and `DefinitionScheduleUpdated`. `ActionSetCreated` and `DefinitionCreated` event payloads gain a `schedule` field; absent payloads fall back to the column default for replays of older events.
* `CreateActionSet` and `CreateDefinition` handlers now accept and persist schedule. Two new RPC handlers `UpdateActionSetSchedule` and `UpdateDefinitionSchedule` append the corresponding events. ControlService delegates both. `actionSetToProto` / `definitionToProto` populate the new Schedule field.

### Container-wins resolution (`def > set > action`)

This inverts the current per-action precedence (which was action > set > definition). Operators lose per-action carve-outs against an assigned container, but in exchange every member of an assigned set/definition fires together when the container's schedule fires — that's the load-bearing invariant for the agent's ordering guarantee.

* New SQL queries (`queries/reach.sql`) return per-layer reach data with the assignment mode collapsed (EXCLUDED > UNINSTALL > REQUIRED > AVAILABLE) at each container level for the requested device. Three small queries — one per layer — keep each one simple.
* New `resolution.ResolveDeviceTree` builds the device sync tree from those queries. Sets that are members of any reached definition are absorbed into the definition's group; actions in any reached set are absorbed into that set's group; only the leftover direct-action assignments remain on the standalone layer. Member walking reuses the existing `ListDefinitionMembers` / `ListActionSetMembers` / `GetActionByID` queries; set-member lookups are cached so the same set isn't fetched twice when it's both a def member and reached standalone somewhere upstream.
* `InternalHandler.ProxySyncActions` calls `ResolveDeviceTree` first to emit `grouped_actions`, then folds the existing flat resolver's user-layer + permission-derived TTY actions onto `standalone_actions` (filtering out anything already covered by the tree). Container UNINSTALL forces every member action's `desired_state` to ABSENT; standalone UNINSTALL does the same for the one action.

### Wire shape

`SyncActionsResponse` switches from the old flat `actions` field to the SDK PR #40 `standalone_actions` + `grouped_actions` split. User-layer assignments and TTY actions still flow through the existing flat resolver and ride on standalone_actions; only device-layer goes through the new tree. That keeps the change surface bounded and preserves all existing user/TTY semantics.

## Behavior change to flag

Operators who today directly EXCLUDE a single action from an otherwise-Required set will see that direct assignment ignored — the set's mode wins. Owner has explicitly accepted this trade-off (single-user deployment).

## Test plan

- [x] `go build ./server/...`
- [x] `go vet ./server/...`
- [x] Unit tests for the tree resolver: standalone, set absorption, definition absorption, container UNINSTALL/EXCLUDED override.
- [x] End-to-end ProxySyncActions tests: action set assignment emits one grouped_actions entry with members in declared sort_order; set-level UNINSTALL forces ABSENT on every member.
- [x] Handler tests for the new UpdateActionSetSchedule / UpdateDefinitionSchedule.
- [x] Migration 012 applies cleanly and existing handler tests still pass.
- [ ] Agent PR consumes the new wire shape (in flight, separate PR).
- [ ] Web PR consumes the new wire shape and surfaces schedule editors (deferred).

Refs manchtools/power-manage-agent#45

## Summary by CodeRabbit

* **New Features**
  * Add schedule support and update endpoints for action sets and definitions (schedules can be created and updated; schedule now surfaced on returned objects).
  * Sync now separates standalone actions from grouped actions in responses.

* **Bug Fixes**
  * Uninstall on an action set forces members to be marked absent.

* **Chores**
  * DB migration adds a schedule column with an 8-hour default.


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PR #93, opened 2026-05-03
feat: per-group maintenance windows (#58)

### LINKED ISSUE #58: feat: per-group maintenance windows, enforced in device-local time
## Problem

Operators want fleet-wide change-freeze windows that gate **all** action dispatch, independent of the per-action cron schedule. Examples:

- Finance devices: no changes applied during month-end week.
- Engineering workstations: no dispatch during 09:00-17:00 local on weekdays.
- Factory-floor servers: only allow dispatch during the night shift.

Today there is no such primitive. Per-action `cron` gets part of the way, but per-action scheduling forces the operator to coordinate a cron expression across every action independently — and does nothing for instant actions or ad-hoc dispatches.

## Approach — maintenance window as a group attribute, enforced in device-local time

Add a `maintenance_window` JSONB column to both `device_groups` and `user_groups` (migrations `00XX_maintenance_windows.sql`). Windows are a positive allowlist — empty means "always allowed", so the feature is opt-in with zero migration impact.

### Window shape

```json
{
  "schedule": [
    {"days": ["mon", "tue", "wed", "thu", "fri"], "allow": "22:00-06:00"},
    {"days": ["sat", "sun"], "allow": "00:00-23:59"}
  ]
}
```

- `days`: any of `mon|tue|wed|thu|fri|sat|sun` (lowercase, three-letter).
- `allow`: `HH:MM-HH:MM`, crossing midnight supported (start > end ⇒ continues into next day).
- Multiple entries combine as **OR** within the window (any entry matching = allowed).

### Local-time enforcement — agent side

Only the agent knows its own `time.Local`. The server cannot do the check because the server sees UTC + timezone metadata at best, and the operator explicitly asked for local-time semantics ("2am local" must mean 2am wherever the device is).

- Server ships the **union** of the device's group windows in each `SyncActionsResponse`. Union (most permissive) is the operator-friendly choice: "allowed if any of my groups says it's OK."
- Agent evaluates `time.Now().Local()` against the union at dispatch time, inside `agent/internal/scheduler/scheduler.go:runDueActions`.
- An action that isn't currently in a permitted window is deferred; the scheduler re-checks on its next tick (existing 1-minute cadence via `DefaultCheckInterval`).

### Instant-action bypass

`ACTION_TYPE_REBOOT`, `ACTION_TYPE_SYNC`, and any future instant actions bypass the window by default. An admin hitting "reboot now" from the UI expects immediate execution. A per-dispatch `respect_maintenance_window: true` flag on the new one-shot scheduling RPC (#57) lets admins opt in when they want a scheduled action to still honor the window.

### Resolution across overlapping groups

A device is in multiple groups (assignment sources). The effective window is the **union** across those groups:

- Device's own group windows (all device groups the device belongs to, directly or via dynamic rules).
- User-group windows from any action assigned via a user-group source targeting this device.

Union = most permissive. Rationale: "allowed if any group says it's OK" matches operator intuition better than "must be OK in every group" (intersection). An operator who wants strict intersection can always collapse devices into fewer groups.

## Test plan

- Store-layer: round-trip the JSONB structure through `group_handler.Set/Get`; reject invalid day strings and time formats.
- Resolution: unit test the union across three overlapping groups; assert expected combined window.
- Agent scheduler: table test — given a time + union window, does `runDueActions` dispatch or defer?
- Integration: fake clock, assign a device to a group with a 22:00-06:00 window, advance fake clock, verify dispatch only fires inside the window.

## Acceptance criteria

- DeviceGroup + UserGroup carry `maintenance_window` JSONB.
- `SyncActionsResponse` includes the resolved union window for the device.
- Agent defers non-instant actions outside the window and retries on next tick.
- Instant actions bypass the window unless explicitly flagged to respect it.
- Web UI edits windows on both group types; display includes a "showing in device-local time" hint.

## Out of scope

- Approval-based override ("force run now anyway") — separate ticket.
- Per-action-type overrides (e.g., "this particular package install ignores windows") — separate ticket if needed.
- Timezone-specific windows (`"Europe/Berlin"` as window tz rather than device-local) — explicitly not what the operator asked for.

PaulDotterer:
## Summary

- Persist `maintenance_window` JSONB on `device_groups_projection` and `user_groups_projection` via two new event types
- New RPCs `SetDeviceGroupMaintenanceWindow` + `SetUserGroupMaintenanceWindow`
- `ProxySyncActions` emits the resolved union of every reaching group's window (device groups + user groups assigned to the device) so the agent can gate `runDueActions` in device-local time
- Validation + union resolution share `sdk/go/maintenance` with the agent so semantics stay bit-for-bit identical
- Empty window = "no constraint"; the migration is opt-in (existing rows default to `'{}'`) and replays produce identical state
- Defensive: `resolveMaintenanceWindowUnion` skips empty rows instead of collapsing the union to "always allowed", so a stale projection or future bypass of the SQL pre-filter cannot accidentally unlock a constrained device

Closes #58

## Test plan

- [x] `go build ./...`
- [x] `go test -run "TestSetDeviceGroupMaintenanceWindow|TestSetUserGroupMaintenanceWindow|TestMaintenance" ./internal/api/`
- [x] Round-trip handler → event → projector → query asserts the window comes back identical
- [x] Negative cases assert validator runs at the handler boundary (bad weekday / zero-length range hit `CodeInvalidArgument`)

## Summary by CodeRabbit

* **New Features**
  * Added API endpoints to set and manage maintenance windows for device groups, including support for multiple schedule entries and clearing windows
  * Added API endpoints to set and manage maintenance windows for user groups
  * Device sync responses now include computed maintenance window information reflecting all applicable constraints

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -107,4 +107,4 @@ require (
 // whatever happens to be in a local ../sdk checkout. Developers who
 // want to iterate against a local SDK override this with a per-dev
 // go.work at their workspace root — see server/README.md for setup.
-replace github.com/manchtools/power-manage/sdk => github.com/manchtools/power-manage-sdk v0.4.1-0.20260503122520-778f885ac65e
+replace github.com/manchtools/power-manage/sdk => github.com/manchtools/power-manage-sdk v0.4.1-0.20260503161545-a3dd8bc45112
--- a/go.sum
+++ b/go.sum
@@ -111,8 +111,8 @@ github.com/lufia/plan9stats v0.0.0-20211012122336-39d0f177ccd0 h1:6E+4a0GO5zZEnZ
 github.com/lufia/plan9stats v0.0.0-20211012122336-39d0f177ccd0/go.mod h1:zJYVVT2jmtg6P3p1VtQj7WsuWi/y4VnjVBn7F8KPB3I=
 github.com/magiconair/properties v1.8.10 h1:s31yESBquKXCV9a/ScB3ESkOjUYYv+X0rg8SYxI99mE=
 github.com/magiconair/properties v1.8.10/go.mod h1:Dhd985XPs7jluiymwWYZ0G4Z61jb3vdS329zhj2hYo0=
-github.com/manchtools/power-manage-sdk v0.4.1-0.20260503122520-778f885ac65e h1:n9MGfcqLOsamGsG5dSsMiytSjywTJVBZXoRkqCT77TI=
-github.com/manchtools/power-manage-sdk v0.4.1-0.20260503122520-778f885ac65e/go.mod h1:HvI/R+FSMZMC4vwuL0Mnf2gcqXyIGJwEK8D7SLY+vh0=
+github.com/manchtools/power-manage-sdk v0.4.1-0.20260503161545-a3dd8bc45112 h1:PMlIH3nqUmyQ+1aa8odJuBZDgtkB1QUd/ZDB+OMiRhI=
+github.com/manchtools/power-manage-sdk v0.4.1-0.20260503161545-a3dd8bc45112/go.mod h1:HvI/R+FSMZMC4vwuL0Mnf2gcqXyIGJwEK8D7SLY+vh0=
 github.com/mattn/go-isatty v0.0.20 h1:xfD0iDuEKnDkl03q4limB+vH+GxLEtL/jb4xVJSWWEY=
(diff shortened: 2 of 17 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

Issue #107, opened 2026-05-03
tracker: Q2 2026 PL/pgSQL → Go listener migration (Phase 1)


PaulDotterer:
## Goal

Move ~30 % of business logic out of PL/pgSQL projector functions into Go event listeners (continuing the pattern set by #77 and #81), retire the duplicate `rebuild_*_projection()` family, and ship a coherent 2026.06 release that pairs the new maintenance-window feature (#58) with structurally reliable search indexing.

## Why

- Today, projection logic lives in `internal/store/migrations/*.sql` as `project_<stream>_event(event events) RETURNS void` PL/pgSQL functions. Reading, debugging, stepping through, and onboarding contributors against PL/pgSQL is materially harder than against Go. That cost compounds with every new projector.
- The same PL/pgSQL surface is what's behind the **stale search results** operators are seeing today: every handler has to remember to call its `enqueueXxxReindex` helper, and the projector functions that update the index are spread across 14 migration files. Centralising in a Go listener (the #77 / #81 pattern) closes both gaps with one architectural move.
- Each ported projector is a small, independently revertable PR. CR review per port stays inside the chill-profile rate limit. No big-bang.
- Sets up the `power-manage-worker` binary (planned for 2026.07) — once projectors are in Go, they can be invoked synchronously (read-your-writes-critical paths) or routed through Asynq into a worker process (background reconcilers, notifications, search reindex).

### Already in flight or shipped

- [x] #58 — per-group maintenance windows (server PR #93 + agent PR #60 in CR loop, web on main)
- [ ] #81 — search indexer as derived projection (the canonical listener pattern; fixes the stale-search bug)

### Infrastructure

- [ ] #94 — replace 13 `rebuild_*_projection()` PL/pgSQL functions with one Go `RebuildAll()`
- [ ] #95 — drop `generate_ulid()` and port `get_*` / `resolve_*` PL/pgSQL helpers to sqlc/Go

### Small projector ports (~half-day each)

- [ ] #96 — `project_security_alert_event` → Go listener
- [ ] #97 — `project_totp_event` → Go listener
- [ ] #98 — `project_lps_password_event` → Go listener
- [ ] #99 — `project_luks_key_event` → Go listener
- [ ] #100 — `project_server_settings_event` → Go listener
- [ ] #101 — `project_role_event` → Go listener
- [ ] #102 — `project_user_role_event` → Go listener
- [ ] #103 — `project_token_event` → Go listener
[...]

PR #109, opened 2026-05-03
refactor: drop get_stream_at + port get_device_sync_interval to sqlc (#95)

### LINKED ISSUE #95: refactor: drop `generate_ulid()` and port `get_*`/`resolve_*` PL/pgSQL helpers to sqlc/Go
## Scope

Retire the small PL/pgSQL helper functions that don't carry business logic — they're either redundant with Go-side code or trivially expressible as sqlc queries.

## Functions to retire

- `generate_ulid()` — fallback for ULID generation. Already covered by `oklog/ulid` in Go; no caller currently relies on the DB version. Drop it; force every caller to provide a ULID from Go.
- `get_stream_at(stream_id, sequence)` — point-in-time stream replay helper. Becomes a sqlc query in `events.sql`.
- `get_device_sync_interval(p_device_id)` — computes effective sync interval (device-level vs group-level minimum). Becomes a sqlc query in `devices.sql` (or a Go helper that calls a simpler query).
- `resolve_inventory_field(p_device_id, field_name)` — extracts a JSONB inventory field. Becomes a sqlc query in `inventory.sql` using the existing JSONB operators.

## Acceptance criteria

- [ ] `generate_ulid()` removed; all callers verified to supply ULIDs from Go (audit `grep -rn "generate_ulid()"`)
- [ ] `get_stream_at` ported to a sqlc query; PL/pgSQL function removed
- [ ] `get_device_sync_interval` ported to sqlc; PL/pgSQL function removed
- [ ] `resolve_inventory_field` ported to sqlc; PL/pgSQL function removed
- [ ] New migration drops all four functions
- [ ] CR review clean

## Tracking

PaulDotterer:
## Summary

Second PR of the Phase 1 PL/pgSQL → Go listener migration (tracker #107). Narrows the original #95 scope from "drop four helpers" to "drop two": `get_stream_at` (no callers) and `get_device_sync_interval` (ported to a sqlc query inlining the same semantics).

The other two (`generate_ulid`, `resolve_inventory_field`) stay because they're called from PL/pgSQL functions inside the dynamic-query interpreter (Group 2, deferred to 2026.07). Removing them without first porting their callers would break dynamic-group evaluation. Migration 016's body documents the deferral so the intent isn't lost.

Closes #95 (with documented deferred follow-up for the remaining two helpers).

## Key design decisions

- **CTE + COALESCE chain** in the new `GetDeviceSyncInterval` query mirrors the PL/pgSQL resolution order exactly: device-level override (non-zero) > smallest non-zero group setting > 0. Behaviour is bit-identical.
- **`NULLIF(x, 0)`** treats 0 as "unset" so the COALESCE chain skips it cleanly. Original PL/pgSQL did the same with `IF NOT NULL AND > 0` chains.
- **MIN of group intervals** matches the original — operator expectation is "join two groups, get the more frequent sync".
- **No public Go API change** — `Queries.GetDeviceSyncInterval` keeps the same `(ctx, deviceID) → int32` signature.

## Test plan

- [x] `go test -run "TestSyncInterval|TestSetDeviceGroupSyncInterval|TestRebuildAll|TestProxySync" ./internal/api/ ./internal/store/` — all pass
- [x] `sqlc generate` produces a clean diff (only the inlined SQL changes)
- [x] `go build ./...` clean
- [x] Down migration restores both functions verbatim — round-trip safe

## Summary by CodeRabbit

* **Refactor**
  * Consolidated device synchronization interval logic in the database layer to compute intervals inline instead of relying on internal helper routines.
* **Chores**
  * Added a migration that removes the now-unused helper routines while providing a rollback path to restore them.

### DIFF
--- a/internal/store/generated/devices.sql.go
+++ b/internal/store/generated/devices.sql.go
@@ -129,12 +129,12 @@ func (q *Queries) GetDeviceByID(ctx context.Context, arg GetDeviceByIDParams) (D
 
 const getDeviceSyncInterval = `-- name: GetDeviceSyncInterval :one
 WITH device_override AS (
-    SELECT NULLIF(sync_interval_minutes, 0) AS interval
+    SELECT CASE WHEN sync_interval_minutes > 0 THEN sync_interval_minutes END AS interval
     FROM devices_projection
     WHERE id = $1::TEXT AND is_deleted = FALSE
 ),
 group_min AS (
-    SELECT MIN(NULLIF(dg.sync_interval_minutes, 0)) AS interval
+    SELECT MIN(CASE WHEN dg.sync_interval_minutes > 0 THEN dg.sync_interval_minutes END) AS interval
     FROM device_groups_projection dg
     JOIN device_group_members_projection dgm ON dgm.group_id = dg.id
     WHERE dgm.device_id = $1::TEXT
--- a/internal/store/generated/devices.sql.go
+++ b/internal/store/generated/devices.sql.go
@@ -128,9 +128,41 @@ func (q *Queries) GetDeviceByID(ctx context.Context, arg GetDeviceByIDParams) (D
 }
 
 const getDeviceSyncInterval = `-- name: GetDeviceSyncInterval :one
-SELECT get_device_sync_interval($1::TEXT) AS sync_interval_minutes
+WITH device_override AS (
+    SELECT NULLIF(sync_interval_minutes, 0) AS interval
+    FROM devices_projection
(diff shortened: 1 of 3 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai full review

Issue #81, opened 2026-04-25
feat: search indexer as a derived projection (sibling of #77, ~48 handler call sites)


PaulDotterer:
## Problem

Same maintenance shape as #77, applied to search indexing. Every handler in `internal/api/` that mutates an indexable entity (user, device, role, group, action, action set, definition, identity provider) has to remember to call its `enqueueXxxReindex` helper, which fans out to `searchIdx.EnqueueReindex / EnqueueRemove / EnqueueMemberAdded / EnqueueMemberRemoved`.

**~48 handler call sites today** (rough grep count). Every new handler / event type adds another site to remember. The 1h `Rebuild` in `cmd/indexer/` is the durability safety net so drift is bounded — but the cognitive cost of "did I remember to enqueue?" stays.

Risk profile is lower than #77 (operator sees stale search results within minutes-to-an-hour rather than silent feature failure), but a privilege-adjacent leak surface still exists — e.g. a soft-deleted user lingering in a search dropdown could let an admin pick them and trigger operations against the deleted account.

## Design — port the rc11 #77 listener pattern

The store-side `EventListener` infrastructure landed in #77. Reuse it.

### 1. Search-affected classifier

```go
// internal/search/listener.go (new)
type SearchOp int
const (
    SearchOpNone SearchOp = iota
    SearchOpReindex          // EnqueueReindex(scope, id, data)
    SearchOpRemove           // EnqueueRemove(scope, id)
    SearchOpMemberAdded      // EnqueueMemberAdded(...)
    SearchOpMemberRemoved    // EnqueueMemberRemoved(...)
)

type SearchAffected struct {
    Op          SearchOp
    Scope       string // "user", "device", "role", "user_group", "device_group", "action", ...
    ID          string
    ParentScope string // for member events
    ParentID    string
    ChildScope  string
    ChildID     string
    ChildName   string
}

func AffectedFromEvent(e store.PersistedEvent) []SearchAffected
```

Multiple results because a single event can affect several index entries (e.g. `UserGroupMemberAdded` reindexes the user AND records a member relationship on the group).

### 2. Listener registration

`cmd/control/main.go` registers the listener post-commit, same pattern as #77:

```go
st.RegisterEventListener(search.NewListener(svc.SearchIdx(), logger.With("component", "search_listener")))
```

### 3. Delete the 48 scattered enqueue calls

[...]

PR #110, opened 2026-05-04
feat: search-index listener infrastructure (Phase 1 of #81)

PaulDotterer:
## Summary

Lands the store-side search listener that classifies AppendEvent post-commit notifications into search-index reindex/remove ops via `api.AffectedSearchOps`. Mirrors the pattern from #77's SystemActionListener.

**Phase 1 scope**: user + device coverage end-to-end. The classifier returns nil for every other stream type today, so the existing handler-side `enqueueXxxReindex` calls remain the source of truth for those scopes during the Phase 1 validation window.

## Why two phases instead of one big bang

- Listener and handler-side enqueues **both fire** today. The Asynq worker dedupes on (scope, id) so functional behaviour is unchanged. If the classifier misses an event type, the handler-side enqueue still fires — zero risk of stale search results during validation.
- **Phase 2** (separate PR) removes the ~48 handler-side enqueues once the classifier has been validated in production. Each follow-up PR adds a scope (device_group, user_group, action_set, definition, action, compliance_policy, execution) and removes its handler-side calls in the same diff — small, reviewable, individually revertable.

This split was a deliberate scope decision documented in the issue body update — it's safer than the one-PR plan in the original issue and keeps every diff under CR's chill-profile review window.

Closes none yet (#81 stays open until Phase 2 lands).

## Test plan

- [x] `TestAffectedSearchOps` table-drives every covered event type plus explicit no-op cases for events that should NOT trigger reindex (UserPasswordChanged, UserSshKeyAdded, UserLoggedIn, DeviceCertRenewed, DeviceAssigned). A future event type added to a handler must show up here with an explicit case so missed coverage is caught at review time.
- [x] `TestSearchListener_NilDepsAreSafe` ensures the factory never panics AppendEvent when search isn't configured.
- [x] `go build ./...` clean
- [x] No regressions on existing search-related handler tests

## Summary by CodeRabbit

* **Improvements**
  * Enhanced search index synchronization through a unified event handling mechanism, improving reliability of search results as data changes.

* **Tests**
  * Added test coverage for search event classification and handling.

### DIFF
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -432,6 +432,18 @@ func main() {
 		searchIdx := search.New(rdb, st, aqClient, logger.With("component", "search"))
 		svc.SetSearchIndex(searchIdx)
 
+		// Phase 1 of #81: register the store-side search listener so
+		// every event that affects the search index funnels through
+		// one classifier (api.AffectedSearchOps) instead of ~48
+		// scattered handler-side enqueueXxxReindex calls.
+		//
+		// During Phase 1 the listener and the existing handler-side
+		// enqueues both fire — the Asynq worker dedupes on
+		// (scope, id) so functional behaviour is unchanged. Phase 2
+		// (separate PR) removes the handler-side calls once we've
+		// validated the listener catches everything in production.
+		st.RegisterEventListener(api.SearchListener(st, searchIdx, logger.With("component", "search_listener")))
+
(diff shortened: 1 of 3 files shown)


PaulDotterer:
@coderabbitai full review

PaulDotterer:
@coderabbitai full review

PR #113, opened 2026-05-04
feat: search-index listener covers action_set + definition (#81 Phase 2c)

PaulDotterer:
## Summary

Phase 2c of #81. Extends `api.AffectedSearchOps` to classify every ActionSet* + Definition* stream event and removes the matching handler-side enqueues. Introduces `cascadeIDsForRemove()` helper so the listener can pull parent IDs from `search.Index.GetReverseMembers` before enqueueing the remove, replacing the inline lookup the delete handlers used to do.

After this lands, the search-index listener owns the lifecycle of action_set + definition reindex/remove. The action↔set and set↔definition membership EDGE updates (`EnqueueMemberAdded`/`EnqueueMemberRemoved`) still fire from handlers — Phase 2c.2 (a follow-up PR) will extend `SearchAffected` with member-edge variants and remove those calls.

## Cleanup details

- **action_set_handler.go**: drops `enqueueSetReindex` helper, 4 call sites, the `GetReverseMembers` + `EnqueueRemove` block in `DeleteActionSet`, and now-unused `search` / `taskqueue` imports.
- **definition_handler.go**: same treatment.

## Race-window note (cascade lookup)

`cascadeIDsForRemove` reads from Redis. The reverse-member entries for a deleted parent persist in Redis until the indexer worker actually processes the `EnqueueRemove` task — so the listener's lookup still finds them in the single-remove flow. For rapid-burst deletes there's a window where the indexer worker has already cleared one parent's reverse set when the listener for the next delete fires. Drift is bounded by the periodic indexer reconciler (~1h), so the worst case is "a parent's denormalised member list is briefly stale."

## Test plan

- [x] `TestAffectedSearchOps` extended with explicit cases for every newly-covered event type
- [x] No regressions on `TestCreateActionSet`, `TestDeleteActionSet`, `TestRenameActionSet`, `TestCreateDefinition`, `TestDeleteDefinition`, `TestRenameDefinition`
- [x] `go build ./...` clean

## Release Notes

* **Refactor**
  * Streamlined search index operations for action sets and definitions through updated event handling.

* **Tests**
  * Extended test coverage for search event classifications and operations.

### DIFF
--- a/internal/api/action_set_handler.go
+++ b/internal/api/action_set_handler.go
@@ -9,10 +9,8 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	pm "github.com/manchtools/power-manage/sdk/gen/go/pm/v1"
-	"github.com/manchtools/power-manage/server/internal/search"
 	"github.com/manchtools/power-manage/server/internal/store"
 	db "github.com/manchtools/power-manage/server/internal/store/generated"
-	"github.com/manchtools/power-manage/server/internal/taskqueue"
 )
 
 // ActionSetHandler handles action set RPCs.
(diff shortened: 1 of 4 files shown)


PaulDotterer:
@coderabbitai full review

PaulDotterer:
@coderabbitai full review

Issue #97, opened 2026-05-03
refactor: port `project_totp_event` projector to Go event listener


PaulDotterer:
## Scope

Port the `project_totp_event(event events) RETURNS void` PL/pgSQL function to a Go event listener following the pattern from #77 (system actions) and #81 (search indexer).

## Function characterisation

tiny — write-once-per-event projection, no member churn.

## Current location

- PL/pgSQL function: `project_totp_event` in `internal/store/migrations/` (find with `grep -l project_totp_event internal/store/migrations/*.sql`)
- Trigger dispatch: `project_event()` → `project_totp_event` for the matching stream type
- Event types projected include: TOTP setup/verification/disable events

## Target

- New Go file: `internal/projectors/totp_projector.go` (or wherever the convention lands after #81)
- Implements `Apply(ctx, tx, event)` for every event type the PL/pgSQL function handled
- Registered in `cmd/control/main.go` next to the other Go projectors
- Live path: invoked synchronously by the post-commit listener, in the same DB transaction as the event append (read-your-writes consistency preserved)
- Rebuild path: invoked by the unified `RebuildAll` from #94

## Acceptance criteria

- [ ] Go projector handles every event type the PL/pgSQL function did
- [ ] Each event type covered by table-driven tests against testcontainers Postgres
- [ ] PL/pgSQL function deleted in a new migration; `project_event()` trigger no longer dispatches to it
- [ ] Existing handler tests still pass (read-your-writes preserved)
- [ ] CR review clean

## Tracking

Part of the Q2 2026 PL/pgSQL → Go listener migration (Phase 1) tracker.

PR #117, opened 2026-05-04
feat: port project_security_alert_event to Go listener (#96)

### LINKED ISSUE #96: refactor: port `project_security_alert_event` projector to Go event listener
## Scope

Port the `project_security_alert_event(event events) RETURNS void` PL/pgSQL function to a Go event listener following the pattern from #77 (system actions) and #81 (search indexer).

## Function characterisation

small (~50 lines), already half-touched by the #77 listener landing.

## Current location

- PL/pgSQL function: `project_security_alert_event` in `internal/store/migrations/` (find with `grep -l project_security_alert_event internal/store/migrations/*.sql`)
- Trigger dispatch: `project_event()` → `project_security_alert_event` for the matching stream type
- Event types projected include: `SecurityAlertCreated`

## Target

- New Go file: `internal/projectors/security_alert_projector.go` (or wherever the convention lands after #81)
- Implements `Apply(ctx, tx, event)` for every event type the PL/pgSQL function handled
- Registered in `cmd/control/main.go` next to the other Go projectors
- Live path: invoked synchronously by the post-commit listener, in the same DB transaction as the event append (read-your-writes consistency preserved)
- Rebuild path: invoked by the unified `RebuildAll` from #94

## Acceptance criteria

- [ ] Go projector handles every event type the PL/pgSQL function did
- [ ] Each event type covered by table-driven tests against testcontainers Postgres
- [ ] PL/pgSQL function deleted in a new migration; `project_event()` trigger no longer dispatches to it
- [ ] Existing handler tests still pass (read-your-writes preserved)
- [ ] CR review clean

## Tracking

PaulDotterer:
## Summary

**Canary for the projector-migration pattern.** First scope ported under the agreed approach: a pure event→projection function lives in `internal/projectors/`, and a post-commit listener applies it to the database. The handler can call the same pure function on a freshly-appended event to render its API response immediately while the listener writes the row asynchronously — both produce identical state because they're the same function on the same event.

If this canary feels right, we apply the same shape to #97–#106 (totp, lps_password, luks_key, server_settings, role, user_role, token, identity_provider, scim_group_mapping, user_selection).

Closes #96. Unblocks #97–#106.

## Design summary

- **Pure function** (`SecurityAlertProjectionFromEvent`, `SecurityAlertAckParamsFromEvent`): no I/O, deterministic, fully unit-testable without a database. Returns `ErrIgnoredEvent` for events outside the scope so the listener wrapper can no-op silently.
- **Post-commit listener** (`SecurityAlertListener`): calls the pure functions, writes via two new sqlc queries (`InsertSecurityAlertProjection` / `AcknowledgeSecurityAlertProjection`). Errors logged + swallowed per the post-commit notification contract.
- **Idempotent insert** via `ON CONFLICT (event_id) DO NOTHING` so a crash-recovered re-fire is safe.
- **Out-of-order ack** (acknowledge arrives before the alert insert reaches the projection) is logged with the alert_id — same operator-visibility contract the deleted PL/pgSQL projector had via `RAISE EXCEPTION` + sidecar trigger's `EXCEPTION` handler.
- **Migration 017** drops `project_security_alert_event`, `project_security_alert_trigger`, and the `AFTER INSERT ON events` trigger. Down restores all three verbatim.

## Why async is safe for `security_alert` specifically

- The `InboxWorker` that emits `SecurityAlert` events is itself an async Asynq task — no synchronous caller waits on the projection write.
- No API handler reads back a security_alert immediately after emitting one; the web UI's list query naturally tolerates the small post-commit gap.

The handler-side immediate-render pattern (the killer feature of the pure-function split) isn't exercised by this scope, but it's available the moment a synchronous handler emits a `SecurityAlert` and wants to render the alert in its response. Subsequent ports (#97–#106) will exercise this path because their handlers are synchronous RPC paths.

## Test plan

- [x] `TestSecurityAlertProjectionFromEvent_Pure` — 5 cases (happy path, missing details, wrong event_type, wrong stream_type, malformed JSON)
- [x] `TestSecurityAlertAckParamsFromEvent_Pure` — happy + malformed UUID
- [x] `TestSecurityAlertListener_EndToEnd` — real Postgres via testcontainers, AppendEvent → poll for projection row → assert contents
- [x] `go build ./...` clean
- [x] No regressions on existing tests

## Summary by CodeRabbit

* **Refactor**
  * Migrated security alert projection handling from database-level triggers to application-level event listeners, improving system performance and maintainability.

* **Tests**
  * Added comprehensive test coverage for security alert event projection and listener functionality, including edge cases and end-to-end scenarios.

### DIFF
--- a/cmd/control/main.go
+++ b/cmd/control/main.go
@@ -365,6 +365,26 @@ func main() {
 	// 3) Periodic reconciler — durability safety net for the listener,
 	//    catches any event whose effect on system actions the
 	//    listener doesn't yet know about. Default 1m.
+	// Phase 1 of the PL/pgSQL → Go projector migration (#96): the
+	// security_alert projector now lives in Go via a post-commit
+	// listener. The (now-deleted) PL/pgSQL function
+	// project_security_alert_event + its sidecar trigger ran inside
+	// the AppendEvent transaction. The Go listener fires post-commit,
+	// but no consumer of security_alerts_projection relies on
+	// read-your-writes after an alert append, so the async write is
+	// safe. ON CONFLICT (event_id) DO NOTHING in the insert keeps it
+	// idempotent against post-crash re-fires.
+	//
+	// Registered OUTSIDE the SystemActions guard below — the
+	// projector listener has no SystemActions dependency, and a
+	// deployment with SystemActions disabled must still update the
+	// security_alerts_projection (the PL/pgSQL trigger this replaces
+	// always fired regardless of subsystem availability).
+	st.RegisterEventListener(projectors.SecurityAlertListener(
+		st,
+		logger.With("component", "security_alert_projector"),
+	))
+
 	if svc.SystemActions() != nil {
 		// (1) Startup sweep — keeps the existing Info line so
 		// operators see the one-shot convergence in boot logs.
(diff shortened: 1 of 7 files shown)


PaulDotterer:
@coderabbitai full review

PR #122, opened 2026-05-05
feat: port project_server_settings_event to Go listener (#100)

### LINKED ISSUE #100: refactor: port `project_server_settings_event` projector to Go event listener
## Scope

Port the `project_server_settings_event(event events) RETURNS void` PL/pgSQL function to a Go event listener following the pattern from #77 (system actions) and #81 (search indexer).

## Function characterisation

trivial key/value upsert.

## Current location

- PL/pgSQL function: `project_server_settings_event` in `internal/store/migrations/` (find with `grep -l project_server_settings_event internal/store/migrations/*.sql`)
- Trigger dispatch: `project_event()` → `project_server_settings_event` for the matching stream type
- Event types projected include: Server-wide settings events

## Target

- New Go file: `internal/projectors/server_settings_projector.go` (or wherever the convention lands after #81)
- Implements `Apply(ctx, tx, event)` for every event type the PL/pgSQL function handled
- Registered in `cmd/control/main.go` next to the other Go projectors
- Live path: invoked synchronously by the post-commit listener, in the same DB transaction as the event append (read-your-writes consistency preserved)
- Rebuild path: invoked by the unified `RebuildAll` from #94

## Acceptance criteria

- [ ] Go projector handles every event type the PL/pgSQL function did
- [ ] Each event type covered by table-driven tests against testcontainers Postgres
- [ ] PL/pgSQL function deleted in a new migration; `project_event()` trigger no longer dispatches to it
- [ ] Existing handler tests still pass (read-your-writes preserved)
- [ ] CR review clean

## Tracking

PaulDotterer:
## Summary

Fifth port under the projector-migration pattern. Replaces `project_server_settings_event` PL/pgSQL with `projectors.ServerSettingsListener`.

- Pure decoder `ServerSettingsUpdatedFromEvent` returns `*bool` fields so omitted-vs-explicit-false is distinguishable. Empty payload is NOT an error — mirrors the PL/pgSQL `COALESCE(NULL, existing)` no-op semantics.
- Listener factory wraps a single `UpdateServerSettings` call (no `store.WithTx` since it's one statement).
- New sqlc query uses `sqlc.narg()` to get nullable `*bool` params; COALESCE preserves columns when the listener passes nil.
- New `WHERE projection_version < $N` guard rejects stale reconciler replays — tightens the PL/pgSQL version which stamped `projection_version` without a guard.
- Migration replaces the PL/pgSQL function with a no-op stub (the shared `project_event()` trigger still dispatches; final cleanup after #106 drops the dispatcher itself).
- Wired via `projectors.WireAll`.

## Read-your-writes preserved

`settings_handler.UpdateServerSettings` reads back from the projection AFTER `AppendEvent` returns. `fireListeners` is synchronous (project memory `feedback_post_commit_listener_is_sync.md`), so the read sees the listener's write.

## Test plan

Tests written FIRST per saved TDD feedback.

- [x] Pure decoder: happy path with both fields, partial updates (each field omitted individually), empty payload returns nil-only struct, wrong stream/event type → `ErrIgnoredEvent`, malformed payload bytes
- [x] Integration: full update, partial update preserves omitted field via COALESCE, `projection_version` guard rejects stale replay, wrong-stream-type ignored
- [x] All tests pass locally (`go test ./server/internal/projectors/...`)

Refs tracker #107. Closes #100.

## Summary by CodeRabbit

* **Refactor**
  * Moved server settings projection from a database stored procedure into the application to centralize update processing; user behavior unchanged.
* **Chores**
  * Added a safe migration to disable the old DB projector and new conditional update logic to the application flow.
* **Tests**
  * Added end-to-end and unit tests to validate full, partial, stale-replay, and wrong-stream handling.

### DIFF
--- a/internal/projectors/server_settings.go
+++ b/internal/projectors/server_settings.go
@@ -54,11 +54,11 @@ func ServerSettingsUpdatedFromEvent(e store.PersistedEvent) (ServerSettingsUpdat
 	return p, nil
 }
 
-// ApplyServerSettingsUpdateForTest exposes the SQL UPDATE side of
-// the listener so tests can drive it with a synthetic
-// projection_version (e.g. an artificially-stale replay). Production
-// code goes through the listener; this is test-only seam.
-func ApplyServerSettingsUpdateForTest(ctx context.Context, st *store.Store, u ServerSettingsUpdate) error {
+// ApplyServerSettingsUpdate runs the COALESCE-based UPDATE behind
+// the listener. Shared between the production listener and the test
+// suite, which uses it to drive a synthetic projection_version (e.g.
+// an artificially-stale replay) without faking an event.
+func ApplyServerSettingsUpdate(ctx context.Context, st *store.Store, u ServerSettingsUpdate) error {
 	return st.Queries().UpdateServerSettings(ctx, db.UpdateServerSettingsParams{
 		UserProvisioningEnabled: u.UserProvisioningEnabled,
 		SshAccessForAll:         u.SshAccessForAll,
--- a/internal/projectors/server_settings.go
+++ b/internal/projectors/server_settings.go
@@ -1,0 +1,68 @@
+package projectors
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/manchtools/power-manage/server/internal/store"
+	db "github.com/manchtools/power-manage/server/internal/store/generated"
+)
+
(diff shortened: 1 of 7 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PR #129, opened 2026-05-05
feat: port project_user_selection_event to Go listener (#106)

PaulDotterer:
## Summary

**Eleventh and FINAL port** under tracker #107's projector-migration pattern. Replaces `project_user_selection_event` PL/pgSQL with `projectors.UserSelectionListener`.

Single event type (`UserSelectionChanged`), single UPSERT statement — no transaction wrap needed. `ON CONFLICT (device_id, source_type, source_id) DO UPDATE` matches the PL/pgSQL projector — repeated selection toggles refresh the row in place.

Tightening: the conflict-update path now has a `WHERE projection_version < EXCLUDED.projection_version` guard rejecting stale reconciler replays. The PL/pgSQL projector lacked it.

> ⚠️ **Known issue (#125):** the `user_selections` entry in `AllRebuildTargets` will dispatch to the no-op stub after this PR merges, breaking emergency rebuild via `RebuildAll(ctx, "user_selections")`. Same pattern affects `roles` + `tokens` (already broken since their respective ports merged). The fix is to dispatch via the Go listener pipeline in `runOneTarget`; filed separately at #125 — recommend tackling that BEFORE the cleanup migration after this PR merges, so the cleanup migration can also drop the affected `AllRebuildTargets` entries that point at no-op stubs.

## Test plan

Tests written FIRST per saved TDD feedback.

- [x] Pure decoder: happy paths, default `selected=false` on missing, required-field validation per variant, wrong stream/event type, malformed payload
- [x] Integration: Select→Deselect (UPSERT in place, no duplicate row)
- [x] Integration: per-source-type scoping (`user` vs `user_group` don't collapse into one row)
- [x] Integration: `projection_version` stale-replay guard
- [x] Integration: wrong-stream-type ignored
- [x] All tests pass locally (`go test ./server/internal/projectors/...`)

## Tracker #107 progress

After #106 merges: **11 / 11 ports complete.** Ready for the cleanup migration + the un-ported-domain-projector future tracker.

Refs tracker #107. Closes #106.

## Summary by CodeRabbit

* **New Features**
  * Add user selection projection to record selections per device/source with per-source scoping and stable IDs.
  * Upsert behavior prevents stale updates from overwriting newer selection states.

* **Chores**
  * Projection wiring updated to register the new listener; database projector disabled in migration (app-level projector used).

* **Tests**
  * Added tests covering event validation, upsert logic, scoping, and stale-replay protection.

PR #130, opened 2026-05-05
chore: server error-handling + audit-log sweep

PaulDotterer:
## Summary

Seven related fixes from the recent audit + sibling sweeps. All independent of the in-flight projector ports — touches different files than #129.

1. **GetDeviceLpsPasswords + GetDeviceLuksKeys** (HIGH per audit): 6 bare \`fmt.Errorf\` returns replaced with \`apiErrorCtx\`. Clients now receive structured \`ErrorDetail\` instead of opaque wrapped-error strings. Covers all error paths: GetCurrent / GetHistory / Decrypt × 2 in each handler.

2. **CreateUser**: drop the misleading "any AppendEvent failure → ErrEmailAlreadyExists" branch. \`Store.AppendEvent\`'s internal 23505-retry loop masks the original \`PgError\`, so distinguishing a true unique violation from a transient failure is unreliable here. Surfaces structured Internal error + logs the actual cause. (A future caller-level pre-check before AppendEvent could restore the AlreadyExists code; deferred.)

3. **CreateUser**: \`GetRoleByName("User")\` failure escalated from silent \`if err == nil\` to a logged Error so operators can triage "user has no roles" cases without waiting for user tickets.

4. **CreateUser**: \`UserRoleAssigned\` append failures escalated from Warn to Error with explicit "user will be missing this role" wording.

5. **role_handler.go:359**: fixed wrong error constant — was returning \`ErrCannotRenameSystemRole\` ("System roles cannot be modified") for the "cannot remove last user from Admin role" precondition. New \`ErrCannotRemoveLastAdmin\` constant added in errors.go.

6. **UserLoggedIn audit gaps**: append failures across \`totp_handler\`, \`sso_handler\`, \`auth_handler\` escalated from Warn to Error with "AUDIT GAP" prefix. RPCs still succeed (tokens already minted, no safe rollback) but the gap is loud — matches the security framing that an attacker triggering event-store exhaustion could otherwise perform unlogged logins.

7. **Dead code**: \`hashesAny\` in \`totp_handler.go\` was allocated from \`hashes\` but never used; the event payload uses \`hashes\` directly.

## Cross-repo dependency (must merge first)

\`TestErrorCodeParityWithTSSDK\` reads \`sdk/ts/errors.ts\` via filesystem. CI must check out SDK at a commit that includes the matching const for parity to pass.

- **SDK**: https://github.com/manchtools/power-manage-sdk/pull/45 (one-line addition: \`ErrCannotRemoveLastAdmin\`)
- **Web**: \`dc84f65\` (already pushed direct-to-main per project workflow — adds \`error_cannot_remove_last_admin\` paraglide keys for en + de)

## Test plan

- [x] \`go build ./server/...\` clean
- [x] \`TestErrorCodeParityWithTSSDK\` passes locally (workspace SDK has the const)
- [x] Targeted: \`TestCreateUser\`, \`TestCreateRole\`, \`TestRevokeRole\` all pass
- [x] No file overlap with PR #129; safe to land in either order

Refs the audit findings list (Critical & High device_handler.go:632, plus 6 mediums + 1 dead code).

## Summary by CodeRabbit

* **Bug Fixes**
  * Audit append failures during login (password, SSO, TOTP) now log as explicit errors ("AUDIT GAP") but no longer block login.
  * Device password/key retrieval and decryption failures now return standardized API errors.

* **Improvements**
  * Clearer error when removing the last Admin user.
  * User creation: early duplicate-email check, clearer failure handling, and auto-assignment of the built-in User role when none provided.
* **Privacy**
  * Login logs no longer include user email.

### DIFF
--- a/internal/api/auth_handler.go
+++ b/internal/api/auth_handler.go
@@ -126,7 +126,7 @@ func (h *AuthHandler) Login(ctx context.Context, req *connect.Request[pm.LoginRe
 		)
 	}
 
-	h.logger.Info("user logged in", "user_id", user.ID, "email", user.Email)
+	h.logger.Info("user logged in", "user_id", user.ID)
 
 	protoUser := userToProto(user)
 	// Populate user roles
--- a/internal/api/auth_handler.go
+++ b/internal/api/auth_handler.go
@@ -115,7 +115,8 @@ func (h *AuthHandler) Login(ctx context.Context, req *connect.Request[pm.LoginRe
 		ActorType:  "user",
 		ActorID:    user.ID,
 	}); err != nil {
-		h.logger.Warn("failed to append UserLoggedIn event", "user_id", user.ID, "error", err)
+		h.logger.Error("AUDIT GAP: failed to append UserLoggedIn event; password login proceeded without audit record",
+			"user_id", user.ID, "error", err)
 	} else {
 		h.logger.Debug("event appended",
 			"request_id", middleware.RequestIDFromContext(ctx),
--- a/internal/api/device_handler.go
+++ b/internal/api/device_handler.go
@@ -629,13 +629,13 @@ func (h *DeviceHandler) GetDeviceLpsPasswords(ctx context.Context, req *connect.
 	// Get current passwords
 	current, err := h.store.Queries().GetCurrentLpsPasswords(ctx, req.Msg.DeviceId)
 	if err != nil {
-		return nil, fmt.Errorf("get current LPS passwords: %w", err)
+		return nil, apiErrorCtx(ctx, ErrInternal, connect.CodeInternal, "failed to load current LPS passwords")
 	}
 
 	// Get password history
 	history, err := h.store.Queries().GetLpsPasswordHistory(ctx, req.Msg.DeviceId)
 	if err != nil {
-		return nil, fmt.Errorf("get LPS password history: %w", err)
+		return nil, apiErrorCtx(ctx, ErrInternal, connect.CodeInternal, "failed to load LPS password history")
 	}
 
 	// Convert to proto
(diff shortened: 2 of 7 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
Two outside-diff Major findings raised this round on user_handler.go (lines 286-304 and 140-175). Both are real, but both are pre-existing in code this PR didn't introduce — applying my own audit-defer discipline:

**Line 286-304 (ListAllInheritedRoles N+page perf):** real perf bug, but the fix needs a new sqlc query (`ListInheritedRolesByUserIDs`) — out-of-scope for an apiErrorCtx + audit-log sweep. Filing as a separate issue + handling in a focused perf PR.

**Line 140-175 (partial-write on UserRoleAssigned failure):** I deliberately escalated the log from Warn to Error in this PR but kept the RPC succeeding. The reviewer's framing ("never silently ignore errors" per coding guidelines) is correct in principle, but the realistic options are:
  - Fail the RPC after `UserCreated` is already committed → orphaned user in DB + client retries to AlreadyExists confusion
  - Make `UserCreated` + role-assignment atomic → architecture change (event-sourcing makes this hard; would need either combining the events into one or a transactional outbox)

Neither is a sweep-PR-shaped change. The current state (loud Error log + return 200) is the least-bad of bad options pending the architectural fix. The `slog.Error` escalation IS the audit-discipline fix here — operators now see this in journalctl ERROR not in Warn. Filing the atomicity question as a follow-up tracker.

Inline finding (line 86-88 — pre-check error swallowing) addressed in the next push.

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai resume

PaulDotterer:
@coderabbitai review

PR #123, opened 2026-05-05
feat: port project_role_event to Go listener (#101)

PaulDotterer:
## Summary

Sixth port under the projector-migration pattern. Replaces `project_role_event` PL/pgSQL with `projectors.RoleListener`.

Three event types:
- `RoleCreated` — INSERT … ON CONFLICT DO NOTHING (replay-safe)
- `RoleUpdated` — partial UPDATE; pointer fields on the decoded payload distinguish "field present" from "field omitted". Empty-string Name is collapsed to nil at the listener layer to match PL/pgSQL's `NULLIF(name, '')` semantics. COALESCE preserves omitted fields. New `WHERE projection_version < $N` guard rejects stale reconciler replays.
- `RoleDeleted` — `is_deleted = TRUE` + cascade DELETE on `user_roles_projection`, wrapped in `store.WithTx` so the projection never observes "role marked deleted but memberships remain".

## Read-your-writes preserved

`role_handler.{Create,Update}Role` read back from `roles_projection` AFTER `AppendEvent` returns. `fireListeners` is synchronous (project memory `feedback_post_commit_listener_is_sync.md`), so the read sees the listener's write.

## Test plan

Tests written FIRST per saved TDD feedback.

- [x] Pure decoders: happy paths, defaults (description/permissions/is_system), partial-update semantics (each field omitted/empty/explicit-empty), `name=""` → preserved-via-NULLIF, wrong stream/event type → `ErrIgnoredEvent`, malformed bytes
- [x] Integration: Create→Update lifecycle with partial updates (description-only, permissions-clear), Delete cascade to `user_roles_projection`, `projection_version` stale-replay guard, wrong-stream-type ignored
- [x] All tests pass locally (`go test ./server/internal/projectors/...`)

Refs tracker #107. Closes #101.

## Summary by CodeRabbit

* **Refactor**
  * Role projection logic moved from DB triggers to an application-level projector; DB migration added to disable the old in-DB projection.

* **New Features**
  * App-side role listener registered to handle create/update/delete events with validation and preserved partial-update semantics.
  * Projection operations now use version guards and support safe insert/update/soft-delete plus cascade deletion of role memberships.

* **Tests**
  * Added integration test ensuring stale delete replays do not remove role memberships.

PR #131, opened 2026-05-06
fix: route ported projectors through Go applier in RebuildAll (#125)

### LINKED ISSUE #125: RebuildAll dispatches to no-op PL/pgSQL stubs for Go-ported projectors
## Bug

\`Store.RebuildAll\` (\`internal/store/rebuild.go\`) dispatches each event via \`SELECT project_<X>_event(events.*)\`. After tracker #107 ports a projector to a Go listener, the corresponding PL/pgSQL function becomes a no-op stub. So:

1. RebuildAll TRUNCATEs the projection table.
2. RebuildAll iterates every event and calls the no-op stub.
3. The projection stays empty.

## Currently affected (already merged)

- \`roles\` target → \`project_role_event\` (no-op since #101) → wipes \`roles_projection\` on rebuild.

## Will be affected by remaining #107 ports

- \`tokens\` target — already in registry; will break when #103 token port merges.
- \`user_selections\` target — already in registry; will break when #106 user_selection port merges.

## Not affected (no rebuild target was ever added)

- security_alert, totp, lps_password, luks_key, server_settings, user_role, identity_provider, scim_group_mapping, user_group_mapping, scim_etc.

## Fix options

A. **Dispatch ported stream types through the Go listener pipeline.** Modify \`runOneTarget\` to detect ported stream types and call \`fireListeners\` against synthesised \`PersistedEvent\`s instead of the PL/pgSQL function. Cleanest semantically (replays use the same code path as live events).

B. **Keep PL/pgSQL projector body alongside the no-op stub for rebuild only.** Move the original PL/pgSQL body to a separate \`rebuild_<X>_projection_event(event)\` function that RebuildAll dispatches to, leaving the no-op stub for live trigger dispatch. Maintains two implementations of the same logic — risk of drift.

C. **Drop ported entries from \`AllRebuildTargets\`.** Operators lose the ability to emergency-rebuild Go-ported projections via this path. Cheapest fix; least functional.

Recommend (A) — single source of truth in Go, no PL/pgSQL drift surface.

## How this slipped through

CR caught it on PR #124 (#102 user_role) but with imprecise framing. Verified the orphan function \`rebuild_user_roles_projection\` it mentioned doesn't exist — #94 dropped all 13 \`rebuild_*_projection()\` functions before tracker #107 began.

The real bug was introduced by #101 (role port) and is dormant until an operator runs \`RebuildAll(ctx, "roles")\`.

## Tracking

Follow-up to tracker #107.

PaulDotterer:
## Summary

Closes manchtools/power-manage-server#125.

After tracker #107 ported eleven projectors from PL/pgSQL to Go listeners, the legacy `RebuildAll` dispatcher kept calling the `project_<X>_event()` PL/pgSQL functions — now no-op stubs — for every event matching a target's stream types. The result for the three ported projectors that own a `rebuildTarget` (`roles`, `tokens`, `user_selections`) was that emergency rebuilds TRUNCATEd the projection and then dispatched every replayed event to a no-op, leaving the projection permanently empty.

This change wires a Go-side applier registry into the rebuild pipeline:

- `store.RebuildApply` + `Store.RegisterRebuildApply` — new boot-time registry that maps a target name to a Go closure operating on tx-bound queries.
- `runOneTarget` — split into `dispatchViaGoApplier` (used when a Go applier is registered) and `dispatchViaPlpgsql` (legacy path for not-yet-ported targets).
- `projectors.{ApplyRole,ApplyToken,ApplyUserSelection}` — exposed as the transactional cores of the three affected listeners. The listener wrappers now delegate to them, and `WireAll` registers them under the matching `AllRebuildTargets` names.
- The asymmetric-guard discipline introduced in CR #123 for `ApplyRole.RoleDeleted` (skip the cascade DELETE when the version-guarded SoftDelete affected zero rows) is preserved — the listener still wraps in `WithTx` for atomicity and the rebuild path inherits atomicity from the outer rebuild transaction.

This unblocks the cleanup migration that drops the eleven PL/pgSQL stubs and the `project_event()` dispatcher.

## Test plan

- [x] New `TestRebuildAll_PortedRole_RoundTrip` — proves `RebuildAll(ctx, "roles")` restores the projection after a TRUNCATE. Fails red without the fix (no rows in result set), passes green with it.
- [x] New `TestRebuildAll_PortedToken_RoundTrip` — same shape for the tokens target.
- [x] New `TestRebuildAll_PortedUserSelection_RoundTrip` — same shape for the user_selections target.
- [x] Existing `TestRebuildAll_RoundTripsThroughEventStore` (users, not yet ported) still passes — fallback path unchanged.
- [x] Existing `TestRebuildAll_NoArgsRebuildsEverything` still passes — every registered target rebuilds.
- [x] Existing `TestRebuildAll_UnknownTargetIsRejected` still passes — selectTargets validation unchanged.
- [x] Existing role/token/user_selection listener tests pass — confirms the listener-side refactor is behaviour-preserving (asymmetric guard, ignored-event handling, payload-decode errors).
- [x] `go build ./...` clean.

## Summary by CodeRabbit

* **Refactor**
  * Centralized event processing for roles, tokens, and user selections with transactional handling and consolidated error propagation.
* **New Features**
  * Added a Go-based rebuild pathway with registration and wiring for projector targets as an alternative to DB-side replay.
* **Tests**
  * Added end-to-end rebuild round-trip tests for roles, tokens, and user selections.

### DIFF
--- a/internal/projectors/role_listener.go
+++ b/internal/projectors/role_listener.go
@@ -30,28 +30,66 @@ func RoleListener(st *store.Store, logger *slog.Logger) store.EventListener {
 		if e.StreamType != "role" {
 			return
 		}
-		switch e.EventType {
-		case "RoleCreated":
-			applyRoleCreated(ctx, st, logger, e)
-		case "RoleUpdated":
-			applyRoleUpdated(ctx, st, logger, e)
-		case "RoleDeleted":
-			applyRoleDeleted(ctx, st, logger, e)
+		// RoleDeleted needs the SoftDelete + cascade DELETE wrapped
+		// in a transaction so the projection never observes "role
+		// marked deleted but memberships remain". The other event
+		// types are single statements and run on the autocommit
+		// connection. ApplyRole handles all three when called with
+		// tx-bound queries (the rebuild path), so we share its body
+		// for RoleDeleted via WithTx and short-circuit the simple
+		// cases through the pool.
+		if e.EventType == "RoleDeleted" {
+			err := st.WithTx(ctx, func(q *store.Queries) error {
+				return ApplyRole(ctx, q, e)
+			})
+			if err != nil {
+				logger.Warn("role projector: failed to apply RoleDeleted",
+					"event_id", e.ID, "role_id", e.StreamID, "error", err)
+			}
+			return
+		}
+		if err := ApplyRole(ctx, st.Queries(), e); err != nil {
+			logger.Warn("role projector: failed to apply event",
+				"event_id", e.ID, "event_type", e.EventType, "role_id", e.StreamID, "error", err)
 		}
 	}
 }
 
-func applyRoleCreated(ctx context.Context, st *store.Store, logger *slog.Logger, e store.PersistedEvent) {
+// ApplyRole is the transactional core of the role projector. The
+// listener wraps it for live-event dispatch (using WithTx for
+// RoleDeleted's two-write atomicity); the rebuild path
+// (manchtools/power-manage-server#125) registers it via
+// RegisterRebuildApply so RebuildAll re-derives the projection from
+// the event store instead of dispatching to the no-op PL/pgSQL stub.
+//
+// The asymmetric-guard discipline for RoleDeleted is preserved:
+// when the version-guarded SoftDelete affects zero rows, the cascade
+// DELETE is skipped — a stale RoleDeleted re-applied later must not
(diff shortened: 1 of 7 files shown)


PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PaulDotterer:
@coderabbitai review

PR #134, opened 2026-05-06
perf: scope inherited-roles query to the user-list page (drop unscoped variant)

PaulDotterer:
## Summary

\`ListUsers\` populated each page user's inherited roles by calling \`ListAllInheritedRoles\`, which scanned every row of \`user_group_members_projection × user_group_roles_projection × roles_projection × user_groups_projection\` — regardless of how many users were on the page. The handler then filtered down to the page in Go. Cost grew linearly with system-wide group membership count on every paginated request, with no benefit for the common case of small page sizes.

Replace the unscoped query with \`ListInheritedRolesByUserIDs\`, which takes the page's user IDs as \`TEXT[]\` and filters at the database layer. The handler now passes \`pageUserIDs\` (collected during the existing user→proto conversion) and the database returns only the rows it cares about.

Drop the unscoped \`ListAllInheritedRoles\` entirely — leaving it around invites a future caller to reach for it and reopen the perf bug.

## Test plan

- [x] \`go build ./...\` clean.
- [x] \`go test -run TestListUsers ./internal/api/...\` passes.
- [x] sqlc regen clean (new query generated, old query removed).
- [x] No remaining callers of \`ListAllInheritedRoles\` after sweep.

## Release Notes

* **Performance Improvements**
  * Optimized user list loading to fetch inherited roles more efficiently by processing roles on a per-page basis, resulting in faster page load times when viewing user listings.

### DIFF
--- a/internal/api/user_handler.go
+++ b/internal/api/user_handler.go
@@ -288,28 +288,34 @@ func (h *UserHandler) ListUsers(ctx context.Context, req *connect.Request[pm.Lis
 	nextPageToken := buildNextPageToken(int32(len(users)), offset, pageSize, count)
 
 	protoUsers := make([]*pm.User, len(users))
+	pageUserIDs := make([]string, len(users))
 	for i, u := range users {
 		protoUsers[i] = userToProto(u)
 		h.populateUserRoles(ctx, protoUsers[i])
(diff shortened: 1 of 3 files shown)


PaulDotterer:
@coderabbitai review

PR #159, opened 2026-05-07
refactor: port action_set projector to Go listener (#136)

PaulDotterer:
## Summary

First Phase 2 projector port from tracker [#136](https://github.com/manchtools/power-manage-server/issues/136). Ports \`project_action_set_event\` from PL/pgSQL to a Go listener using the canonical pattern validated by all 11 Phase 1 ports.

The action_set projector is the simplest of Phase 2 (per #136's
ordering recommendation) and the explicit "build pattern muscle"
first port for medium-projector territory.

## What landed

- **Pure decoder** \`internal/projectors/action_set.go\` — handles all
  7 event types (ActionSetCreated, ActionSetRenamed,
  ActionSetDescriptionUpdated, ActionSetMemberAdded,
  ActionSetMemberRemoved, ActionSetMemberReordered, ActionSetDeleted).
  Defaults match the PL/pgSQL projector exactly.
- **Listener** \`internal/projectors/action_set_listener.go\` — uses
  \`st.WithTx\` for the multi-write events (member add/remove/delete
  cascade). Stale-replay guard via \`projection_version\` on every
  UPDATE; \`:execrows\` short-circuit on \`n == 0\` per the
  identity_provider model (avoids the F020/F021 LPS hazard).
- **TDD test file** \`internal/projectors/action_set_test.go\` — 619
  LOC. Per-event decoder tests + a full-lifecycle listener test
  with real Postgres via \`testutil.SetupPostgres\`.
- **Migration** \`030_action_set_projector_to_go.sql\` — no-ops the
  PL/pgSQL \`project_action_set_event(events)\` function. Wrapped
  in \`-- +goose StatementBegin\` / \`-- +goose StatementEnd\` per the
  goose-with-PL/pgSQL rule.
- **WireAll registration** — registers the listener and the rebuild
  applier (\`st.RegisterRebuildApply("action_sets", ApplyActionSet)\`).
  The runOneTarget guard at rebuild.go:286 is the safety net.
- **sqlc queries** \`internal/store/queries/action_sets.sql\` — guarded
  UPDATE/UPSERT statements + the cascade DELETE on definitions
  member_count.

## Test plan

- [ ] Lint workflow (added in PR #141) green
- [ ] Tests workflow green (real Postgres via testcontainers)
- [ ] Rebuild path: \`RebuildAll\` against a fresh DB reproduces the
      same projection state the PL/pgSQL projector would have
      produced.

Part of #136.

## Release Notes

* **Chores**
[...]

Issue #136, opened 2026-05-06
Tracker: port the 11 remaining PL/pgSQL projectors to Go listeners


PaulDotterer:
## Context

Tracker #107 ported eleven projectors from PL/pgSQL to Go listeners (closed via #131 + #132 + #133). Eleven domain projectors still live in PL/pgSQL, dispatched via the trimmed-down \`project_event()\` trigger created by migration 028:

\`\`\`
user, device, action, definition, action_set, device_group,
assignment, execution, user_group, compliance, compliance_policy
\`\`\`

These remain PL/pgSQL because they were not part of the rc11 scope. This tracker is the follow-up that ports them, completes the migration, and removes the \`project_event()\` trigger entirely.

## Pattern (validated by tracker #107)

Each port follows the established projector-migration pattern:

1. **Pure decoder** \`<Stream>FromEvent(e)\` returns a typed payload, returns \`ErrIgnoredEvent\` for unknown event types.
2. **Listener factory** \`<X>Listener(st, logger) store.EventListener\` thin wrapper that calls Apply with \`st.Queries()\` (or \`st.WithTx\` for multi-write atomicity).
3. **Apply core** \`Apply<X>(ctx, q *store.Queries, e store.PersistedEvent) error\` — transactional core, takes tx-bound queries so it can be reused by RebuildAll's Go-applier path.
4. **Listener test file** with TDD-first red → green coverage.
5. **Migration** that no-ops the PL/pgSQL projector (matches the 018-027 shape).
6. **WireAll registration** of the listener AND the rebuild applier (\`st.RegisterRebuildApply(\"<target>\", Apply<X>)\`) for any stream with a \`rebuildTarget\` entry in \`AllRebuildTargets\`.

See \`server/internal/projectors/role_listener.go\` + \`role_test.go\` + migration 022 for the canonical example.

## Per-stream complexity notes

- **user** — has the most complex projector (display_name, given_name, profile updates, linux UID, password hash, TOTP fields, multiple event types). Highest LOC. Sees write traffic from CreateUser, SSO auto-create, SCIM provisioning. Coordinate with the partial-write fix in #135.
- **device** — agent enrollment, label updates, group membership joins. Heavy multi-write paths.
- **action** + **definition** — these share \`project_action_event\` (definition events synthesise action rows for compliance policies). Port together to avoid splitting the dispatch.
- **action_set** — straightforward.
[...]

PR #174, opened 2026-05-08
refactor: port user_group projector to Go listener (Part of #136)

PaulDotterer:
## Summary

- Replaces the PL/pgSQL `project_user_group_event()` function with `projectors.UserGroupListener` — third Phase 2 port under tracker #136 (after action_set in PR #159 and assignment in PR #173).
- Listener handles all ten event types (Created, Updated, QueryUpdated, MaintenanceWindowSet, Deleted, MemberAdded, MemberRemoved, RoleAssigned, RoleRevoked, MembersRebuilt) plus the cascading multi-write paths (Deleted nukes members + role-assignments + scim_group_mappings + dynamic eval queue rows; MembersRebuilt replaces the entire member set).
- Asymmetric-guard discipline matches the role / identity_provider / action_set / assignment ports: every UPDATE on `user_groups_projection` carries an explicit `WHERE projection_version < $N` guard, the guarded soft-delete uses `:execrows`, and the cascade short-circuits when n == 0 so a stale `UserGroupDeleted` re-applied later does not silently nuke a freshly-restored group.
- Multi-write events (MemberAdded/Removed, Deleted, MembersRebuilt, QueryUpdated when flipping to dynamic) wrap their writes in `store.WithTx` so each cascade is internally atomic.
- Migration `033_user_group_projector_to_go.sql` no-ops the PL/pgSQL function.
- `WireAll` registers `UserGroupListener` and `RegisterRebuildApply("user_groups", ApplyUserGroup)` so emergency rebuild dispatches through the Go applier.

## Test plan

- [x] New unit + projector integration tests in `internal/projectors/user_group_test.go` cover decoder defaults, every event type, the soft-delete revival branch, stale-replay guard rejection, and the cascade short-circuit on stale `UserGroupDeleted`.
- [x] Existing `internal/projectors`, `internal/store` (rebuild), and `internal/api` integration suites still pass — `RebuildAll` now dispatches `user_groups` through the Go applier.
- [x] `gofmt -l`, `go vet ./...`, `staticcheck ./...` all green.
- [x] CR-CLI returned 0 findings before push.

## Summary by CodeRabbit

* **Refactor**
  * Moved user-group projection processing to a new Go listener and wiring for more reliable, maintainable event handling.
* **New Features**
  * Added robust handling for dynamic/static group transitions, maintenance windows, membership rebuilds, role assignments, and soft-delete cascades.
* **Bug Fixes**
  * Prevented stale-event replays from overwriting newer state; ensured idempotent membership/role updates and correct member counts.
* **Tests**
  * Added extensive unit and end-to-end tests covering decoders, listener behavior, replay-guarding, cascades, and dynamic-group semantics.


PaulDotterer:
Addressed both critical findings (1+2) by adding a single `ClaimUserGroupForMembership :execrows` guard query that runs BEFORE any membership-table mutation. The bump+claim is atomic with the conditions check (parent exists, not dynamic, not deleted, version is newer). Recount via live `COUNT(*)` so ON CONFLICT DO NOTHING idempotency can't drift the counter. Two new regression tests (`StaleMemberAdded`, `StaleMembersRebuilt`) lock the fixed behaviour.

Filed the SCIM rebuild blast-radius (#3) as #175 — that needs an `ApplySCIMGroupMapping` refactor and a new rebuild target ordering, which is outside this port's scope. The PL/pgSQL function on main has the same bug; the port carried it over verbatim per the in-code comment.

PR #326, opened 2026-06-06
feat(deploy): cut over from redis-stack-server to valkey-bundle 9.1.0

### LINKED ISSUE #319: feat(deploy): migrate from redis-stack to valkey + valkey-search 1.2+
The deploy stack currently uses `docker.io/redis/redis-stack-server:7.4.0-v0` (compose.yml:87) because valkey-search 1.0.x was missing TEXT field support and `FT.DROPINDEX DD`. Both gaps are now manageable:

- **TEXT fields**: resolved upstream in [valkey-search 1.2.0](https://github.com/valkey-io/valkey-search/releases/tag/1.2.0) (released 2026-03-17). Issue [valkey-io/valkey-search#801](https://github.com/valkey-io/valkey-search/issues/801) closed.
- **`FT.DROPINDEX DD`**: still open as [valkey-io/valkey-search#590](https://github.com/valkey-io/valkey-search/issues/590), but **we don't depend on it**. `internal/search/index.go:106` already calls `FT.DROPINDEX` without the `DD` flag and runs a `SCAN`-based cleanup pass for the indexed hashes (`search:*`, `reverse:*`, `members:*`). That pattern works on both backends.

The codebase's `FT.CREATE` schemas use TEXT + TAG + NUMERIC + SORTABLE (per `internal/search/index.go`), all of which valkey-search 1.2.0 supports.

## Scope

- `deploy/compose.yml` — replace the `valkey` service image with valkey core + valkey-search. Options:
  - Prebuilt: `valkey/valkey-bundle` (ships the module loaded by default) — simplest, but verify it bundles 1.2+.
  - Manual: `valkey/valkey:9.0.x` + `--loadmodule /path/to/valkey-search.so` via the `valkey.conf` template.
- `deploy/valkey.conf.template` — update the `loadmodule` path from `/opt/redis-stack/lib/redisearch.so` to the valkey-search path on whatever image we settle on.
- `deploy/setup.sh render_valkey_config()` — no change needed; the renderer is path-agnostic.
- Healthcheck (if any) — verify the existing `redis-cli` health probe still works against valkey (it should — `redis-cli` is reused as `valkey-cli` and both protocols are wire-compatible).
- Existing data dir compat — verify a `data/valkey` from redis-stack can be read by valkey, OR document that the migration requires a one-time reindex (we already have `RebuildSearchIndex` RPC + `FlushSearchData()` for this).

## Out of scope for this issue

- Migrating existing fielded data (we'll just trigger `RebuildSearchIndex` on first boot after the cutover).
- Switching `task signing key` / Asynq queue storage backends — Asynq works over plain Valkey/Redis, no change needed there.
- Switching the in-memory token store / TOTP cache — same as above.

## Risk + sequencing

Prep the change locally, soak on a staging install for at least a week, then bundle as the headline change for 2026.07-rc1. The cutover is reversible by reverting the image tag.

Background: surfaced during 2026.06-rc2 testing — see [`project_image_tag_contract.md`](/home/pdotterer/.claude/projects/-home-pdotterer-Schreibtisch-Projekte-power-manage/memory/project_image_tag_contract.md) for the wider deploy story.

PaulDotterer:
Closes manchtools/power-manage-server#319.

The compose stack was pinned to `redis/redis-stack-server:7.4.0-v0` because earlier valkey-search builds were missing TEXT field support (and `FT.DROPINDEX DD`, though we don't depend on it). Both gaps are resolved upstream in valkey-search 1.2+, which ships in `valkey/valkey-bundle:9.1.0` (and `unstable`). This PR cuts over so we drop the Redis Inc. dependency on redis-stack.

## Summary

- `deploy/compose.yml` — image swap to `docker.io/valkey/valkey-bundle:9.1.0`. `command` flips to `valkey-server` (the bundle entrypoint's auto-loadmodule branch only fires on the canonical name, not the `redis-server` symlink). Healthcheck flips to `valkey-cli ping`. Comments refreshed.
- `deploy/valkey.conf.template` — dropped the explicit `loadmodule /opt/redis-stack/lib/redisearch.so` line. The bundle entrypoint (`bundle-docker-entrypoint.sh`) auto-discovers every `.so` under `/usr/lib/valkey/` and appends them as `--loadmodule` args at start. Comments rewritten to document that behaviour.
- `deploy/setup.sh` — comment correction (uid 999 is now the `valkey` user, not the `redis` user).
- `internal/search/index.go` — broadened the `EnsureIndexes` idempotency check from `Index already exists` (RediSearch) to `already exists` so it covers valkey-search 1.2's verbose `Index <name> in database 0 already exists.` wording. The drop-index handler at `index.go:107` and `search_handler.go:194` already cover both backends via the existing `not found` substring; verified empirically against the bundle.
- `internal/search/indexer_test.go` — testcontainer image swap to `valkey/valkey-bundle:9.1.0`. **Test logic unchanged** — the existing suite passes against the new container.
- `internal/terminal/valkey_backend.go` — stale `redis-stack-server` reference in a doc comment updated to call out both backends.

## Why bundle 9.1.0 specifically?

The `9.0.3` tag ships valkey-search **1.0.0** (module ver `10000`), which still doesn't support TEXT fields. The `9.1.0` tag (2026-05-27) and `unstable` both ship valkey-search **1.2** (module ver `66048`). 9.1.0 is the lowest tagged stable bundle that satisfies #319's requirement.

Sibling-swept the codebase for other RediSearch error-string matches:
- `internal/search/index.go:107` (drop) — `"not found"` substring already covers valkey-search 1.2 (`Index with name 'X' not found in database 0`). No change needed.
- `internal/api/search_handler.go:194` (search) — same.

## Validation

- [x] `go build ./...` clean.
- [x] `internal/search` package suite green against `valkey-bundle:9.1.0` (`go test -count=1 ./internal/search/...` → 62s, all green).
- [x] Search-touching `internal/api` tests green (`go test -run 'Search|search' ./internal/api/...` → 80s, all green).
- [x] `internal/terminal`, `internal/gateway/...`, `internal/handler` all green.
- [x] Local compose smoke: `docker compose up -d valkey` healthy in ~8s; modules load (json/search/bloom/ldap); `FT.CREATE idx:smoke ... SCHEMA name TEXT created_at NUMERIC SORTABLE` succeeds via `valkey-cli ping` with the bind-mounted password file.
- [x] `cr review` returned no findings.
- [ ] Soak on staging for ≥1 week before tagging 2026.07-rc1 (per #319's plan).

## Migration notes for operators

- mTLS certs, control / gateway configs, Postgres data — **all unaffected**.
- Existing `deploy/data/valkey/` AOF + RDB carry over. Both images use UID 999 for their service user, so no host-side chown is required.
- On first boot under the new image, the search-index schema is identical (we already use the TEXT+TAG+NUMERIC+SORTABLE shapes valkey-search 1.2 supports). The `EnsureIndexes` idempotency path is the only behaviour difference between backends, and it's covered. For an abundance-of-caution rebuild, hit **Settings → Search → Rebuild index** in the web UI after the cutover.

## Rollback

Revert the image tag in `compose.yml` to `redis/redis-stack-server:7.4.0-v0` and the command to `["redis-server", "/etc/valkey/valkey.conf"]`. The data directory remains compatible. The `internal/search/index.go` substring broadening is forward- and backward-compatible (RediSearch's `Index already exists` still contains `already exists`), so the code change does not need reverting.

## Test plan

- [ ] Single-node staging install soak (≥1 week).
- [ ] Multi-host staging install soak (≥1 week, multiple gateways against the same control + valkey).
- [ ] `RebuildSearchIndex` RPC on a populated install — verify it completes cleanly and all 10 indexes report `last_indexing_finished_time` post-rebuild.
- [ ] Operator-facing search box hits every index scope and returns expected results.

## Release Notes

* **Chores**
  * Migrated underlying infrastructure from Redis Stack Server 7.4.0 to Valkey 9.1.0 for improved performance and maintainability.
  * Updated Go dependencies for enhanced security and stability.
  * Updated configuration templates to align with new deployment architecture.

* **Bug Fixes**
  * Improved error handling in search index creation to support multiple search backend variants with better error detection.

### DIFF
--- a/deploy/compose.yml
+++ b/deploy/compose.yml
@@ -82,9 +82,12 @@ services:
 
   valkey:
     # Pinned to avoid silent upstream upgrades. Bump deliberately when
-    # you've verified the RediSearch module + data format still work
-    # for your deployment.
-    image: docker.io/redis/redis-stack-server:7.4.0-v0
+    # you've verified the valkey-search module + data format still
+    # work for your deployment. The bundle image carries valkey-server
+    # plus the search / json / bloom / ldap modules under
+    # /usr/lib/valkey/; its entrypoint auto-loads any .so it finds
+    # there that isn't already loaded by the config file.
+    image: docker.io/valkey/valkey-bundle:9.1.0
     container_name: pm-valkey
     restart: unless-stopped
     volumes:
@@ -99,10 +102,14 @@ services:
     # /proc/<pid>/cmdline. Now the password is in /etc/valkey/valkey.conf
     # (file permissions enforced by the bind-mount), and the health
     # check authenticates via REDISCLI_AUTH so the cmdline of the
-    # one-shot redis-cli probe stays clean too.
-    command: ["redis-server", "/etc/valkey/valkey.conf"]
+    # one-shot valkey-cli probe stays clean too.
+    #
+    # Command must invoke `valkey-server` (not its `redis-server`
+    # symlink) so the bundle's entrypoint takes the module-autoload
+    # branch — see bundle-docker-entrypoint.sh.
+    command: ["valkey-server", "/etc/valkey/valkey.conf"]
     healthcheck:
-      test: ["CMD-SHELL", "REDISCLI_AUTH=\"$$VALKEY_PASSWORD\" redis-cli ping"]
+      test: ["CMD-SHELL", "REDISCLI_AUTH=\"$$VALKEY_PASSWORD\" valkey-cli ping"]
       interval: 5s
       timeout: 3s
       retries: 3
--- a/deploy/setup.sh
+++ b/deploy/setup.sh
@@ -751,8 +751,8 @@ render_valkey_config() {
     local tmp
     tmp="$(mktemp "${rendered}.tmp.XXXXXX")"
     printf '%s' "$rendered_contents" > "$tmp"
-    # 0644 keeps the file readable by the redis user inside the
-    # container (uid 999 in the redis-stack-server image) without
+    # 0644 keeps the file readable by the valkey user inside the
+    # container (uid 999 in the valkey-bundle image) without
     # needing the operator to chown to a specific uid. The bind-
     # mount carries `ro,z` so the container cannot tamper, and the
     # password is already on disk in .env (same protection level),
--- a/deploy/valkey.conf.template
+++ b/deploy/valkey.conf.template
@@ -1,4 +1,4 @@
-# Valkey/Redis-stack server configuration (audit F-03).
+# Valkey server configuration (audit F-03).
 #
 # Rendered to ./valkey.conf at setup time by setup.sh — the only
 # value substituted in is ${VALKEY_PASSWORD}. The rendered file is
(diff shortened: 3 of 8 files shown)


PaulDotterer:
@coderabbitai review

Issue #336, opened 2026-06-10
build(sqlc): schema overlay for DO-block-hidden scope columns (scope-read SELECTs can't be generated)


PaulDotterer:
## Problem

Migration `010_role_grant_scope_7.sql` adds `scope_kind` / `scope_id` to `user_roles_projection` and `user_group_roles_projection` **inside a `DO $$ … END $$;` PL/pgSQL block** (chosen for atomic multi-statement application). sqlc's schema parser does **not** read `DO` blocks, so it never learns those columns exist.

Consequence:
- sqlc is lenient on INSERT/DELETE column *references* (it binds them as params), so S2's write queries (#334) generate fine.
- But any **SELECT that outputs** `scope_kind` / `scope_id` fails `sqlc generate` with `column "scope_kind" does not exist` (sqlc must resolve the output column's type from the catalog).

This already forced `GetUserScopedGrants` (#7 S2b, PR #335) to be **hand-written** in `internal/store/generated/roles.sql.go` with a `NOTE` documenting the sync contract. Every future scope-read query (the S5 grant listing, the S6 per-scope reconciler cohort reads) will hit the same wall and need hand-maintenance.

Separately, `sqlc generate` at the pinned version reformats all ~27 generated files (version drift), so it can't be run cleanly to regen one query anyway.

## Proposed fix

Add a **plain-DDL sqlc schema overlay** that declares the scope columns for sqlc's parser only (never applied to a DB), and add it to `sqlc.yaml`'s `schema` list:

```yaml
schema: ["migrations/", "sqlc_schema_overlay.sql"]
```

```sql
-- sqlc_schema_overlay.sql — sqlc-parse-only; NOT a goose migration.
-- Re-declares columns that live inside DO-blocks in migrations/ so
-- sqlc can resolve them as query outputs. Keep in sync with 010.
ALTER TABLE user_roles_projection       ADD COLUMN scope_kind TEXT, ADD COLUMN scope_id TEXT;
ALTER TABLE user_group_roles_projection ADD COLUMN scope_kind TEXT, ADD COLUMN scope_id TEXT;
```

Then `GetUserScopedGrants` (and future scope-read queries) generate normally; drop the hand-written method + NOTE.

(The generated-file reformat-churn from the sqlc version drift is a separate, pre-existing concern — worth pinning the sqlc version in CI as a follow-up so `sqlc generate` produces a minimal diff.)

## Acceptance
[...]

Issue #7, opened 2026-04-03
Scoped Access — device group-based RBAC for delegated administration


PaulDotterer:
## Summary

Role assignments can be optionally scoped to a device group. A scoped user only sees and manages devices, actions, assignments, and executions within their scope. User management remains admin-only (unscoped).

## How it works

- Add optional `scope_device_group_id` to `user_roles_projection` — `NULL` = global (current behavior)
- When assigning a role, an optional "Scope to device group" picker narrows the role's reach
- Dynamic device groups work as scopes — automatic scoping as devices match the query

## What scoping affects

| Operation | Unscoped (admin) | Scoped to device group |
|-----------|-----------------|----------------------|
| List/view devices | All | Only devices in scoped group |
| List/view actions | All | Only actions assigned to scoped group |
| Create/delete assignments | Any → any | Visible action → scoped device/group |
| Dispatch actions | Any device | Scoped devices only |
| View executions | All | Scoped devices only |
| View compliance | All | Scoped devices only |
| User management | Yes | No (admin-only, unscoped) |

## Object visibility

Actions/sets/definitions are visible to a scoped user if they are **assigned to** the user's scoped device group. Assignments are the bridge — no labels or tags on objects needed.

## Role escalation prevention

- Scoped users can only assign roles with equal or narrower scope
- Scoped users cannot assign unscoped (global) roles
- Only unscoped admins can create/modify scopes on role assignments

## DB changes

- Add `scope_device_group_id TEXT` (nullable) to `user_roles_projection`
- Update `GetUserPermissions` query to return scope alongside permissions
- JWT: add `scoped_permissions` alongside flat `permissions` list

## Also requires changes in

- `manchtools/power-manage-web` — Role assignment scope picker, filtered list views

PaulDotterer:
V1 scope enforcement is complete and shipped for 2026.07: device/user/terminal scope (#406/#407/#408) + object visibility & device/user Search scope (#462). The sole remaining V2 item (`:assigned` + device-group-scope composition, narrowing-only) is tracked in #463. Closing the epic for the release.

PR #337, opened 2026-06-10
feat: #7 S5 — scope-aware role assignment + revocation (AssignRole/RevokeRole scope tuple, escalation prevention)

PaulDotterer:
Fourth slice of manchtools/power-manage-server#7 — the **grant-authoring** half: `AssignRoleTo{User,UserGroup}` and `RevokeRoleFrom{User,UserGroup}` now consume the `(scope_kind, scope_id)` tuple. **Server-only** — the proto fields, `RoleGrantScopeKind` enum, `AssignRoleScope` permission, `PermissionTargetKind`, and the scope-aware projector writes all already existed (S1/#333, S2/#334); the handlers ignored scope until now. Builds on S2b (#335)'s `UserContext.ScopedGrants`.

## Assign validation (`validateAssignGrantScope`, shared by both handlers)
- **paired-or-neither** — `scope_kind`/`scope_id` set together or not at all;
- **`AssignRoleScope` gate** for any scoped grant;
- **scope-group existence** (device_group / user_group);
- **per-role `target_kind` match** — every permission in the role must accept the scope kind (device_group scope ⇒ all perms `TargetDevice`). Derived self-discoveringly from the permission registry (`TargetKindFor`), **not a hardcoded list**, so a new permission can't silently slip the check.

## Escalation prevention (auth, over the actor's own `ScopedGrants`)
- **`EnforceGrantScopeAuthority`** — a scope-limited admin may attach only a scope its own `AssignRoleScope` authority covers (global ⇒ any; scoped ⇒ equal-id match; sub-group "narrower" containment is a documented future refinement).
- **`EnforceUnscopedGrantAuthority`** — a scope-limited admin may not mint an unscoped, fleet-wide grant.

In V1 `AssignRoleScope` is org-tier (held only by unscoped admins), so the bounded path isn't reachable yet — this is defense-in-depth ahead of per-scope admins.

## Revoke
Targets a SPECIFIC `(subject, role, scope)` grant via new NULL-aware existence checks (`UserHasScopedRole` / `UserGroupHasScopedRole`, hand-written like S2's scope queries — sqlc can't read migration 010's `DO`-block columns, see #336). Revoking a grant that exists only at a **different** scope is rejected (`FailedPrecondition`) rather than silently no-op'd; nothing-there stays an idempotent no-op.

## Test fidelity fix
`AdminContext` now carries every admin permission as an **unscoped grant**, mirroring production — where the flat permission list is *derived* from grants — so the scope-authority checks see an org admin as globally unrestricted rather than as having no scope authority. (Additive; existing tests unaffected.)

## Tests
- **auth units**: target-kind match (incl. cross-kind + TargetUnspecified rejection), both escalation bounds.
- **handler integration** (user + user-group paths): scoped grant end-to-end, `AssignRoleScope` gate, target_kind mismatch, missing scope group, paired-or-neither, escalation out-of-scope + unscoped denial, scope-targeted revoke + wrong-scope rejection.
- `go vet ./...`, `gofmt`, `go build`, `internal/auth`, `internal/projectors`, all targeted scope tests green. Full `internal/api`: **312 pass / 0 fail** (local wall-clock exceeded the default timeout under heavy back-to-back container churn; CI's `-timeout 20m` fresh runner ran the identical-size suite in 562s at S2b). Local CodeRabbit: no findings.

## Follow-ups
The **enforcement-consuming** half — per-device/-user handlers calling `EnforceDeviceScope`/`EnforceUserScope` with a `ScopeResolver` wired to the projections (S6), the per-scope reconciler, and the TTY-gated web scope picker (S9).

Closes part of manchtools/power-manage-server#7.

## Summary by CodeRabbit

* **New Features**
  * Scoped role assignments and revocations: roles can be granted/revoked limited to specific device or user group scopes; assignment/revoke operations are scope-aware and include scope details in audit events.
  * Stronger validation and authority checks to prevent invalid or unauthorized scoped grants.

* **Tests**
  * Expanded end-to-end and unit tests covering scoped grant/revoke semantics, authority enforcement, and scope-edge cases.

### DIFF
--- a/internal/api/role_handler.go
+++ b/internal/api/role_handler.go
@@ -442,8 +442,7 @@ func (h *RoleHandler) RevokeRoleFromUser(ctx context.Context, req *connect.Reque
 	} else {
 		// The targeted grant doesn't exist. If the role IS assigned at a
 		// different scope, the caller targeted the wrong grant — surface
-		// that rather than silently no-op (#7 S5). If nothing exists at
-		// all, fall through as an idempotent no-op.
+		// that rather than silently no-op (#7 S5).
 		hasAny, err := h.store.Repos().Role.UserHasRole(ctx, req.Msg.UserId, req.Msg.RoleId)
 		if err != nil {
 			return nil, apiErrorCtx(ctx, ErrInternal, connect.CodeInternal, "failed to check role assignment")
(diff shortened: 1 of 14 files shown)


PaulDotterer:
@coderabbitai full review

PR #367, opened 2026-06-11
fix(api): centralized last-admin lockout protection (#365)

### LINKED ISSUE #365: security(audit): no privilege ceiling on role/group management + incomplete last-admin protection (High)
From `SECURITY_AUDIT_2026-06-10.md`, finding #8 (High, conditional).

**Privilege escalation:** `CreateRole`/`UpdateRole` validate only that permission keys are *valid* (`ValidPermissionKeys()`), not that the caller *holds* them; `AssignRoleToUser`/`AssignRoleToUserGroup` perform no subset check; `AddUserToGroup` adds arbitrary users to a group with no check on the group's privilege level. So any delegated holder of these RPCs can grant permissions they lack — escalating to full Admin. There is no "grant only what you hold" ceiling anywhere.

**Last-admin lockout:** last-admin protection exists only on `RevokeRoleFromUser`, and it naively counts users with the Admin role (not *enabled* admins, ignores group-inherited admins). `DeleteUser`, `SetUserDisabled`, and stripping Admin perms via `UpdateRole` all bypass it — the sole administrator can be locked out.

## Fix (per maintainer decision)
- **Strict "grant only what you hold":** the caller must hold every permission they grant/assign across CreateRole/UpdateRole/AssignRoleToUser/AssignRoleToUserGroup/AddUserToGroup (holding an unrestricted permission covers its `:self`/`:assigned` scoped forms).
- **Centralized last-admin protection:** a single "would this leave zero enabled admins?" check (counting group-inherited admins) applied across DeleteUser/SetUserDisabled/UpdateRole/RevokeRoleFromUser/group-role-revoke.

PaulDotterer:
Security audit finding **#8 (High)** — part 2 of 2 (the ceiling was #366; this is the last-admin protection).

## The bug
Last-admin protection existed only on `RevokeRoleFromUser`, and it naively counted users with the Admin role (`CountUsersWithRole`) — ignoring whether they're **enabled** and ignoring **group-inherited** admins. `DeleteUser`, `SetUserDisabled`, and stripping the Admin role's permissions via `UpdateRole` all bypassed it, so the sole administrator could be locked out.

## Fix
- `assertOtherEnabledAdminExists`: refuses an op that would leave **zero enabled admins**, where an admin holds the Admin role **directly OR via a group**, filtered to `!Disabled && !IsDeleted`. Wired into `DeleteUser`, `SetUserDisabled` (when disabling), and `RevokeRoleFromUser` (replacing the naive check).
- **Admin system role permission immutability** in `UpdateRole` — closes the strip-Admin-perms lockout (reuses `cannot_remove_last_admin`, no new error code).

## Test harness fix (the reason #8b was split out)
`testutil.CreateTestUser(…,"admin")` previously wrote only the legacy `role` column with **empty `role_ids`**, so it was not a real RBAC admin — with the guard in place it'd see zero admins and (correctly) block every delete/disable. It now creates a genuine `user_roles_projection` assignment via `role_ids`. **Validated the factory change alone passes the full `internal/api` suite (zero blast radius)**, then api + auth pass with the guard.

Red→green tests for each guarded path + Admin-role immutability.

> Remaining narrower vector (noted, follow-up): revoking the Admin role from a **group** that confers it (`RevokeRoleFromUserGroup`). The bootstrap admin is always a *direct* admin, so it can't lock out that way; a correct multi-user guard there needs its own design.

Closes #365.

## Summary by CodeRabbit

* **New Features**
  * Added safeguards to ensure at least one enabled administrator remains in the system.
  * Enhanced validation across admin-related operations including user deletion, account disabling, and role revocation.

* **Tests**
  * Added comprehensive test coverage verifying admin protection mechanisms across multiple operations.

### DIFF
--- a/internal/api/admin_guard.go
+++ b/internal/api/admin_guard.go
@@ -1,0 +1,90 @@
+package api
+
+import (
+	"context"
+
+	"connectrpc.com/connect"
+
+	"github.com/manchtools/power-manage/server/internal/store"
+)
+
+// assertOtherEnabledAdminExists rejects an operation that would remove the LAST
+// enabled administrator. It verifies at least one ENABLED admin OTHER than
+// affectedUserID remains — an admin being a user who holds the Admin system
+// role directly OR via a group. Used by DeleteUser, SetUserDisabled and
+// RevokeRoleFromUser so the sole administrator cannot be locked out (#365).
+//
+// Conservative by design: it requires a *different* enabled admin to remain
+// rather than recomputing the affected user's post-operation admin paths. For a
+// lockout guard, refusing to remove the only OTHER admin is the safe bound, and
+// it counts group-inherited admins (which the prior RevokeRoleFromUser check
+// ignored) and excludes disabled/deleted ones (which it also ignored).
+//
+// LIMITATION (pre-existing): this is a read-side preflight, NOT atomic with the
+// delete/disable/revoke event it guards. Two concurrent admin-removing requests
+// for *different* admins can both observe "another enabled admin exists" and
+// both proceed, racing to zero enabled admins. The prior CountUsersWithRole
+// check had the identical race. The outcome is recoverable — the bootstrap
+// admin (CONTROL_ADMIN_EMAIL/PASSWORD) is re-created on server start — so this
+// strict-improvement guard does not close the race; atomic enforcement in the
+// projection/transaction is a tracked follow-up.
+func assertOtherEnabledAdminExists(ctx context.Context, st *store.Store, affectedUserID string) error {
+	adminRole, err := st.Repos().Role.GetByName(ctx, "Admin")
+	if err != nil {
+		return apiErrorCtx(ctx, ErrInternal, connect.CodeInternal, "failed to resolve Admin role")
+	}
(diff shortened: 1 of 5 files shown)


PaulDotterer:
Thanks @coderabbitai — both addressed:

- **role_handler.go (defer guard until grant confirmed):** fixed. Moved the last-admin check **inside the `hasScoped` block** and kept it gated to unscoped Admin revocations, so a sole group-only admin (or any no-op/scoped revoke) is no longer rejected — the guard now fires only when an unscoped Admin grant will actually be removed. Added `TestRevokeRoleFromUser_ScopedAdminRevokeNotBlocked`.
- **admin_guard.go (make the check atomic):** correct, and pre-existing (the old `CountUsersWithRole` check had the identical TOCTOU race). A proper atomic fix is a projection/transaction-level change (heavy lift) and the outcome is recoverable via the bootstrap admin on restart, so I've documented the limitation in the guard's doc comment and filed https://github.com/manchtools/power-manage-server/issues/369 rather than expand this PR. This PR is a strict improvement over the broken prior check (which ignored enabled-state and group-inherited admins).

PR #389, opened 2026-06-11
feat(security): agent cert revocation via Valkey CRL (#388)

### LINKED ISSUE #388: security(audit): agent cert revocation via Valkey CRL (#6)
SECURITY_AUDIT_2026-06-10.md #6. Agent certs keep 1-year validity, but the gateway never re-checks a presented cert against any revocation state — a leaked, superseded (post-renewal), or deleted-device cert keeps connecting until natural expiry. Design (maintainer-directed): broadcast a CRL via Valkey, no per-connection RPC.

- New `internal/crl` package: a Valkey sorted-set (scored by the revoked cert's expiry → auto-prune, single ZRANGEBYSCORE load) Store (control writes) + in-memory Cache (gateway reads; fail-static on a Valkey blip).
- Control writes the superseded fingerprint on `RenewCertificate` (old fp + old NotAfter) and the device's fingerprint on `DeleteDevice`.
- Gateway caches the CRL on startup + a 30s ticker and checks the presented cert's SHA-256 fingerprint in MTLSMiddleware → 403 if revoked. Shared `ca.FingerprintFromCert` so control-write and gateway-read fingerprints can't drift.
- Keeps 1-year validity (maintainer's call — offline-agent tolerance); revocation is now the enforcement mechanism. (Version-guarded concurrent renewal is a separate, lower-priority item.)

Tests: crl Store/Cache (miniredis, incl. fail-static + expiry prune); RenewCertificate + DeleteDevice revoke; MTLSMiddleware rejects a revoked cert.

PaulDotterer:
Security audit **#6** — agent certs keep their 1-year validity, but the gateway never re-checked a presented cert against any revocation state, so a **leaked, superseded (post-renewal), or deleted-device** cert kept connecting until its natural expiry.

Design (maintainer-directed): broadcast a CRL over the existing **Valkey** infra — **no per-connection RPC**, gateways cache it.

## How it works
- **`internal/crl`** — a Valkey **sorted set** scored by each revoked cert's expiry. That gives a single `ZRANGEBYSCORE` load, automatic age-out (a cert past its own `NotAfter` is rejected by mTLS expiry anyway), and one key. `Store` (control writes) + `Cache` (gateway reads): the cache refreshes on a 30s ticker and **fail-static** on a Valkey blip — it keeps the last good snapshot rather than fail-open to an empty list.
- **Control writes** the superseded fingerprint on `RenewCertificate` (old fp + old `NotAfter` for the TTL) and the device's fingerprint on `DeleteDevice`. Wired via `SetCRLStore` from the Valkey subsystem; nil-safe when no Valkey is configured.
- **Gateway** loads the CRL at startup + refreshes, and rejects a revoked cert in `MTLSMiddleware` (403). New `ca.FingerprintFromCert` is byte-for-byte identical to `FingerprintFromPEM` (hex SHA-256 of DER) so the control-write and gateway-read fingerprints can't drift.

Keeps 1-year validity (your call — offline-agent tolerance); revocation is now the enforcement mechanism. The concurrent-renewal version-guard (the heartbeat-stream-CAS half of #6) is a separate, lower-priority item.

## Tests
- `crl` Store/Cache over miniredis — revoke/load, expiry exclusion + prune, and **fail-static on a refresh error**.
- `RenewCertificate` and `DeleteDevice` put the fingerprint on the CRL.
- `MTLSMiddleware` rejects a revoked agent cert (403, never reaches AgentService) and admits a non-revoked one.

`go vet ./...` caught the changed `MTLSMiddleware` signature's test callers before push.

Closes #388.

### DIFF
--- a/cmd/control/valkey.go
+++ b/cmd/control/valkey.go
@@ -34,6 +34,7 @@ import (
 	"github.com/manchtools/power-manage/server/internal/asynqutil"
 	"github.com/manchtools/power-manage/server/internal/ca"
 	"github.com/manchtools/power-manage/server/internal/control"
+	"github.com/manchtools/power-manage/server/internal/crl"
 	"github.com/manchtools/power-manage/server/internal/gateway/registry"
 	"github.com/manchtools/power-manage/server/internal/search"
 	"github.com/manchtools/power-manage/server/internal/store"
@@ -64,6 +65,11 @@ type valkeySubsystem struct {
 	// one instance so ProxyValidateTerminalToken can validate tokens
 	// minted by the same control replica.
 	TerminalTokenStore *terminal.TokenStore
+
+	// CRLStore is the Valkey-backed certificate revocation list. main() hands
+	// it to the ControlService so renewal/device-deletion revoke the old cert;
+	// gateways read the same Valkey key to enforce it on mTLS connections.
+	CRLStore *crl.Store
 }
 
 // Close stops the Asynq servers and closes the Valkey clients.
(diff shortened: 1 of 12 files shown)


PaulDotterer:
Both addressed @coderabbitai:

- **Critical (FingerprintFromCert nil):** added a nil guard returning "" — defensive on the hot mTLS path. It fails *safe*: a nil cert is already rejected by the peer-class/TLS checks upstream, and an empty fingerprint matches no revoked entry. Added TestFingerprintFromCert_NilSafe.
- **Minor (DeleteDevice lookup error):** now logged (Warn) before falling through best-effort, so a swallowed lookup error can't silently leave a deleted device's cert unrevoked.

Issue #391, opened 2026-06-11
follow-up: close terminal sessions on terminal-grant role revoke (after access recompute)


PaulDotterer:
Follow-up from #390. Terminal-session revocation currently fires on UserDisabled/UserDeleted only. RevokeRoleFromUser was intentionally excluded because a user may still hold terminal access (StartTerminal / TerminalAdmin) via another role — closing on every role-revoke would be wrong. To cover it: on a role-revoke that removed a terminal-granting role, recompute whether the user STILL has terminal access (direct + group roles); only if not, call TerminalHandler.TerminateUserSessions. Reuses the #390 async-listener path.

PR #392, opened 2026-06-11
feat(security): close terminal sessions on user revocation (#390)

### LINKED ISSUE #390: security(audit): terminal sessions survive user revocation (l.174)
SECURITY_AUDIT_2026-06-10.md l.174. `ValidateTerminalToken` runs once at WS accept; a live terminal session then runs with no re-auth, so a user who is **disabled or deleted keeps a root-capable shell** until they disconnect or an admin manually terminates it.

Design (maintainer-directed; refined): the session token is opaque + single-use (GETDEL at accept) so it can't start NEW sessions, and pending tokens expire within the token TTL — the real gap is the ALREADY-OPEN session. Fix: close those on revocation.

- `TerminalHandler.TerminateUserSessions(userID)` — lists active sessions across gateways (ListActiveTerminalSessions, no in-method auth gate) and terminates each via the existing fan-out under a synthetic *system* actor (so the audit event attributes correctly).
- `TerminalRevocationListener` — a post-commit listener on UserDisabled / UserDeleted that runs the fan-out on a **background goroutine** (async, best-effort) so it never blocks the disable/delete RPC. Wired in the Valkey subsystem.
- Scope: disable + delete only (maintainer's call). Role-revoke excluded — a user may keep terminal access via another role; needs a 'still has access?' recompute → follow-up.

Tests: the listener closes sessions on UserDisabled + UserDeleted, and not on re-enable or non-user events (the gateway fan-out itself is covered by the terminal admin RPC tests).

PaulDotterer:
Security audit **l.174** — terminal sessions survive revocation. `ValidateTerminalToken` runs once at WS accept, so a live session then runs with **no re-auth**: a user who is **disabled or deleted keeps a root-capable shell** until they disconnect or an admin manually kills it.

Design (your direction, refined): the token is opaque + **single-use** (`GETDEL` at accept) so it can't start new sessions, and pending tokens expire within the TTL — the real gap is the **already-open** session. Close it on revocation.

## How it works
- **`TerminalHandler.TerminateUserSessions(userID)`** — lists active sessions across gateways (`ListActiveTerminalSessions`, which has no in-method auth gate) and terminates each via the existing fan-out under a synthetic **system** actor, so the audit event attributes correctly.
- **`TerminalRevocationListener`** — a post-commit listener on `UserDisabled` / `UserDeleted` that runs the fan-out on a **background goroutine** (async, best-effort, your call) so it never blocks the disable/delete RPC and survives that request's context being cancelled when the RPC returns. Wired into the Valkey subsystem.
- **Scope:** disable + delete only (your call). Role-revoke is excluded — a user may keep terminal access via another role, so it needs a "still has access?" recompute → follow-up **#391**.

## Tests
The listener closes sessions on `UserDisabled` and `UserDeleted`, and **not** on re-enable or non-user events (via a terminator interface — the gateway fan-out itself is covered by the existing terminal admin RPC tests). `go vet ./...` clean.

Closes #390.

### DIFF
--- a/cmd/control/valkey.go
+++ b/cmd/control/valkey.go
@@ -159,6 +159,13 @@ func newValkeySubsystem(ctx context.Context, cfg *Config, st *store.Store, svc *
 		_ = err
 	}
 	svc.SetTerminalHandler(termHandler)
+
+	// Close a user's live terminal sessions when their access is revoked
+	// (UserDisabled / UserDeleted) — otherwise a disabled/deleted user keeps a
+	// root-capable shell until they disconnect (audit l.174). Runs the gateway
+	// fan-out in the background so it never blocks the disable/delete.
+	st.RegisterEventListener(api.TerminalRevocationListener(termHandler, logger.With("component", "terminal_revocation")))
+
 	if cfg.TerminalGatewayURL != "" {
 		logger.Info("remote terminal sessions enabled",
 			"fallback_gateway_url", cfg.TerminalGatewayURL,
--- a/internal/api/terminal_handler.go
+++ b/internal/api/terminal_handler.go
@@ -678,6 +678,45 @@ func (h *TerminalHandler) TerminateTerminalSession(ctx context.Context, req *con
 	return connect.NewResponse(&pm.TerminateTerminalSessionResponse{}), nil
 }
 
+// TerminateUserSessions force-closes every LIVE terminal session belonging to
+// userID across all gateways. It's the revocation path (audit l.174): a disabled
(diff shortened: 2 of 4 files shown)


PaulDotterer:
Both addressed @coderabbitai — and the panic-recovery one is important: an unrecovered panic in a spawned goroutine crashes the whole control process, which a best-effort session close must never do. Added a defer/recover that logs the panic, plus a completion log (with the context error). New TestTerminalRevocationListener_RecoversPanic pins that a panicking close doesn't take the process down.

PR #397, opened 2026-06-11
fix(security): make last-admin guard atomic with the mutation (#369)

### LINKED ISSUE #369: tech-debt: make last-admin protection atomic with the mutation (close TOCTOU)
Follow-up from #367 (CodeRabbit, #365). The last-admin guard (`assertOtherEnabledAdminExists`) is a read-side preflight, not atomic with the delete/disable/revoke event it guards — two concurrent admin-removing requests for different admins can both pass and race to zero enabled admins. The prior `CountUsersWithRole` check had the identical race; #367 is a strict improvement but does not close it. Fix: enforce the invariant atomically in the user_role/user projection transaction (or serialize admin-removing mutations, e.g. a pg advisory lock around the guard + append). Recoverable today via the bootstrap admin on restart, so Medium.

PaulDotterer:
Follow-up from #367/#365. `assertOtherEnabledAdminExists` was a **read-side preflight, not atomic** with the delete/disable/revoke event it guards — two concurrent admin-removing requests for *different* admins could both observe "another enabled admin exists" and **race to zero**, locking the org out of all admin access (recoverable only via the bootstrap admin on restart).

## Fix
- **`Store.WithAdvisoryLock(ctx, key, fn)`** — a session-level `pg_advisory_lock` taken on a dedicated pooled connection and held across the **entire** callback, including the synchronous post-commit projector write an `AppendEvent` triggers (so the next caller blocks until the prior projection write lands, not just until commit). Explicitly released (a pooled conn isn't closed on Release), detached from ctx so a cancelled request still frees it.
- **`guardedAdminMutation`** runs the guard + the mutating append under that lock. `SetUserDisabled`, `DeleteUser`, and `RevokeRoleFromUser` (unscoped-Admin branch) now go through it. The serialized next caller sees the true count, so the guard refuses the one that would hit zero.

## Tests
- `WithAdvisoryLock` deterministically serializes same-key critical sections.
- Four concurrent admin-disables leave **≥1 enabled admin**. Verified this invariant test goes **red** (races to zero) with the lock bypassed, green with it.

Closes #369.

### DIFF
--- a/internal/api/admin_guard.go
+++ b/internal/api/admin_guard.go
@@ -2,11 +2,43 @@ package api
 
 import (
 	"context"
+	"errors"
 
 	"connectrpc.com/connect"
 
 	"github.com/manchtools/power-manage/server/internal/store"
 )
+
+// advisoryKeyAdminMutation serializes every last-admin-guarded mutation
+// (disable / delete / admin-role-revoke) so the read-side guard and its
+// mutating event append are atomic against a concurrent admin-removing request
+// — closing the TOCTOU where two requests for different admins both observe
+// "another admin exists" and race to zero (#369). The value is "admin" in hex;
+// it only needs to be a stable constant distinct from any other advisory lock.
+const advisoryKeyAdminMutation int64 = 0x61646d696e
+
+// guardedAdminMutation runs the last-admin guard and the mutating append under
+// one advisory lock, so two concurrent admin-removing requests cannot both pass
+// the guard and lock everyone out (#369). affectedUserID is the admin being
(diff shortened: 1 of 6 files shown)


PaulDotterer:
Superseded by #405. The advisory-lock commit (`Store.WithAdvisoryLock` + `guardedAdminMutation`) is included there and extended to cover **all** admin-removing paths — the three user-group demotion paths under the same lock, a self-discovering completeness guard, and the IdP-orphan auto-delete gap that guard surfaced. Closing in favor of #405, which closes #369.

Issue #325, opened 2026-06-05
tracker: sort + filter on list pages — route through valkey-search


PaulDotterer:
## Background

The web UI ships disabled list-page sort buttons (~30 of them) and a handful of empty-relation / TAG filters (e.g. action-sets with no actions, definitions with no sets, compliance policies with no rules, actions unassigned, devices online/offline). Today they're all tooltipped `Coming in 2026.06`. 2026.06 went out without them; we want them in 2026.07 latest, behind the valkey-search cutover (#319).

## Scope decision: extend the search RPC, not the `List*` RPCs

> **Correction to the originally filed scope.** Phase 0.5 of #84 already migrated every UI list page to the scoped `Search` RPC (`apiClient.search(query, SearchScope.XXX, ...)`). The pages do *not* call `ListActions` / `ListDevices` / etc. anymore — they call `Search`. So the original framing of "add sort+filter params to `List*Request` messages and route through valkey-search" was redundant. The work is:
>
> 1. Extend `SearchRequest` with sort params.
> 2. Add the missing TAG / NUMERIC fields to the per-scope index schemas in `internal/search/index.go`.
> 3. Replace the hardcoded `SORTBY <scope-default> DESC` in `search_handler.go` with the request's `sort_field` + `sort_direction`.
> 4. Wire the buttons in the web pages.
>
> `List*` RPCs stay as they are — direct-to-Postgres, default order, no sort. They're for non-UI callers (CI, scripts, hand-rolled clients) that don't need sort/filter and don't want to depend on index freshness. The acceptance criteria for "default page load still hits Postgres" no longer applies — the UI doesn't hit Postgres for lists today, the search RPC is the only path.

Depends on **#319** landing first. Doing the index-schema work against RediSearch and then migrating doubles the indexer schema work in #319; doing it after valkey-search lands means one schema migration, not two.

## Affected scopes — index schema changes

`internal/search/index.go` already declares timestamp fields `SORTABLE` on every index (`created_at`, `updated_at`, `last_seen_at`, `registered_at`, `occurred_at`). What's missing per scope:

| Scope (`SearchScope.*`) | Index | Sortable today | Sortable to add | New TAG / NUMERIC fields to add | Indexer changes |
|---|---|---|---|---|---|
| `ACTIONS` | `idx:actions` | created_at, updated_at | name, type | **`assigned: bool` TAG** — computed from `ActionAssigned` / `AssignmentDeleted` events | new projection slice |
[...]

PR #460, opened 2026-06-24
feat(search): list-page filters via valkey-search (server#325 PR B)

PaulDotterer:
Second slice of #325 — the new-field **filters**, built on the warm↔entityFields consolidation so each field is written once.

## Consolidation first
`refactor(search): route warm* through entityFields` makes `entityFields` the single source of the HSET field shape (8 of 10 warm builders; devices also reuse `EnrichDeviceInventory`). Behavior-preserving (integration green), net −8 LOC.

## Filters
- **assigned** TAG on actions/action_sets/definitions — the entity *itself* directly assigned (`EXISTS assignment WHERE source_id = id`, per the agreed semantic; not its children). Bulk `ListAssignedSourceIDs` per type in warm; per-entity `CountAssignments` incrementally; **reindexed on Assignment Created/Deleted** — the delete event now carries the source tuple (it was empty) so the classifier needs no DB lookup; legacy empty-payload deletes are a no-op (reconcile is the safety net).
- **rule_count** NUMERIC SORTABLE on compliance (+created_at); backs `@rule_count:[0 0]`. **os_name**→TAG (devices), **role** TAG + **last_login_at** (users).
- **Device online/offline**: synthetic `status` filter translated to a query-time `last_seen_at` range (5-min window matching Postgres `ListOffline`).

## Notes / decisions
- `UserRoleChanged` reindexes (role filter correctness); `UserLoggedIn` does **not** (too frequent — last_login is eventually-consistent via reconcile, fine for dormant-account sorting).
- Self-discovering parity guards keep `scopeFilterFields`/`scopeSortableFields` ↔ index in lockstep (caught a missing `last_login_at` filter entry during dev).

## Tests
Unit: parity guards, resolveSort, AffectedSearchOps (assignment→source reindex, compliance/empty no-ops), deviceStatusClause. Integration: search schema+warm green; projector + assignment/search handler tests green (AssignmentDeleted payload change is backward-compatible).

## Summary by CodeRabbit

* **New Features**
  * Search now supports filtering devices by online/offline status and sorting by additional fields such as rule count and last login time.
  * Several search results now include richer metadata like assigned status, user role, and compliance rule counts.

* **Bug Fixes**
  * Deletions now emit more complete event details, improving downstream tracking.
  * Search reindexing now updates more records when assignment or user role changes occur.

* **Refactor**
  * Search indexing and rebuilds now use more consistent data formatting across entities.

### DIFF
--- a/internal/api/assignment_handler.go
+++ b/internal/api/assignment_handler.go
@@ -174,8 +174,10 @@ func (h *AssignmentHandler) DeleteAssignment(ctx context.Context, req *connect.R
 		return nil, err
 	}
 
-	// Verify assignment exists before emitting delete event
-	_, err = h.store.Repos().Assignment.GetByID(ctx, req.Msg.Id)
+	// Verify assignment exists before emitting delete event. Capture the source
+	// tuple so the event carries it (StreamID is the assignment id) — the search
+	// classifier reindexes the source's `assigned` TAG without a DB lookup.
+	assignment, err := h.store.Repos().Assignment.GetByID(ctx, req.Msg.Id)
 	if err != nil {
 		return nil, handleGetError(ctx, err, ErrAssignmentNotFound, "assignment not found")
 	}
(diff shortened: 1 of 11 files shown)


PaulDotterer:
@coderabbitai review

PR #461, opened 2026-06-24
fix(search): NUMERIC empty-relation filters (server#325)

PaulDotterer:
## What

`buildFTQuery` emitted every structured tag filter with TAG syntax (`@field:{v}`), which is a **RediSearch SYNTAX error on a NUMERIC field**. `member_count` and `rule_count` are indexed NUMERIC, so the empty-relation filters could never resolve:

- action-sets with **no actions** (`member_count = 0`)
- definitions with **no sets** (`member_count = 0`)
- compliance policies with **no rules** (`rule_count = 0`)

## Fix

- `IndexSchema.NumericFields()` — derive the NUMERIC-declared fields from the schema (self-discovering, can't drift).
- `numericSearchFields` in the handler — union across all index schemas.
- `buildFTQuery` emits `@field:[v v]` ranges for NUMERIC fields; pipe-separated values OR into `(a|b)`. Non-integer values are dropped before reaching the query string.

## Tests

`TestBuildFTQuery_NumericFieldEmptyRelation`, `_RuleCount`, `_MultipleValues`, `_NonIntegerDropped`. The numeric path is also the live check on `NumericFields()` — if it returned empty, these would fall through to TAG syntax and fail.

Companion web change re-adds the three filter controls these unblock.

Closes the empty-relation gap left open by #460.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search filtering so numeric fields now use valid numeric match syntax instead of tag filters.
  * Non-integer values in numeric filters are ignored, and multiple numeric values are combined correctly.
* **Tests**
  * Added coverage for numeric filter handling, including single values, multiple values, and invalid input.

### DIFF
--- a/internal/api/search_handler.go
+++ b/internal/api/search_handler.go
@@ -440,6 +440,19 @@ func validateFiltersForScopes(ctx context.Context, scopes []string, dateFilters 
 		}
 		return apiErrorCtx(ctx, ErrValidationFailed, connect.CodeInvalidArgument, fmt.Sprintf("filter field %q is only valid for scope=%s", field, strings.Join(supportedBy, ",")))
 	}
+	// A NUMERIC field is emitted as a range (@field:[v v]); a non-integer value
+	// can't be expressed and would silently widen the query (fail-open). Reject
+	// it at the boundary rather than dropping it.
+	for field, value := range tagFilters {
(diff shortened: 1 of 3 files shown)


PaulDotterer:
@coderabbitai review

PR #462, opened 2026-06-26
Scoped access (#7): object visibility + device/user Search scope + event-driven dynamic membership

PaulDotterer:
Closes the object-catalog and device/user Search leaks for scope-restricted admins (server#7). Two commits on this branch.

## 1. Scoped object visibility — actions / action-sets / definitions / compliance policies
All four object permissions were `TargetUnspecified`, so a device-group-scoped operator could `Search`, `Get`, and **mutate** the entire org catalog (every action's `ShellParams` script, every policy — the whole security posture).

- `auth.ObjectScopeListFilter` — caller scope = union of device-/user-group ids from the **JWT** grants (no DB round-trip).
- Server-only `scope_group_ids` TAG on the four object indexes = the object's direct group-target assignments; populated in warm + incremental reindex. Fully fresh (changes only on the object's own assignment events).
- Search appends `@scope_group_ids:{caller groups}` per object scope when restricted.
- 28 Get/mutation handlers: `Get` uses **effective** scope (object + containers, live) → `NotFound` + WARN log out-of-scope (no existence leak); mutations use **direct** scope → `PermissionDenied` (transitive visibility never grants write).
- Self-discovering parity test ties handler enforcement to index filtering both ways.

## 2. Device/user Search scope + event-driven dynamic group membership
The `Search` RPC applied no device/user scope, and the device/user list pages use `Search` (not the scoped `ListDevices`/`ListUsers`) → whole-org leak.

- `scope_group_ids` on `idx:devices` / `idx:users` = the entity's group memberships (same `is_deleted`-excluded source as the auth resolver). Search keys devices off `DeviceScopeListFilter("ListDevices")`, users off `UserScopeListFilter("ListUsers")`.
- **Event-driven dynamic membership (no audit/state drift, no ticker):** the dynamic-group evaluator wrote `*_group_members` directly with no events. It now emits a `*GroupMembersReevaluated` delta event and stops writing directly; a projector applies it (`ClaimDynamic*GroupForMembership` version+dynamic guard → insert/delete → recount). The event is the **source of truth** — a projection rebuild replays it, like static membership. The search listener reindexes the affected devices/users off the same event.

## Verification
`go vet`, `staticcheck`, and the `dyngroupeval` / `projectors` / `store` (rebuild) / `search` / full `api` suites all green. The existing evaluator membership-correctness tests run through the new event path unchanged (they are the integration test). ADR `0024-scoped-object-visibility.md`. No proto/SDK change.

Known fail-closed boundary: transitive-only / individual-device-assigned objects under-show in `Search` (readable via `Get`).

## Summary by CodeRabbit

* **New Features**
  * Search now respects scoped visibility, so results are filtered to items the caller can access.
  * Read access can include container-derived visibility, while edit/delete access requires direct assignment.
  * Dynamic device and user group membership changes are now reflected more consistently in search results.

* **Bug Fixes**
  * Out-of-scope items now appear as not found when viewed, and unauthorized edits return permission denied.
  * Search now avoids leaking items outside the caller’s scope.

### DIFF
--- a/docs/adr/0024-scoped-object-visibility.md
+++ b/docs/adr/0024-scoped-object-visibility.md
@@ -92,12 +92,40 @@ build.
 - **Known boundary (documented, fail-closed):** transitive-only and
   individual-device/user-assigned objects under-show in `Search` for scoped
   callers; they remain reachable via `Get`.
-- **Out of scope of this ADR — a related, broader leak:** the `Search` RPC also
-  applies *no* device/user scope today, so the device/user **list pages** (which
-  use `Search`, not the scoped `ListDevices`/`ListUsers` RPCs) leak the whole org
-  to a scoped admin. Closing it needs the same `scope_group_ids` mechanism on the
-  device/user indexes plus a membership-change reindex cascade — and, for
-  *dynamic* groups (whose re-evaluation emits a **bulk** `*MembersRebuilt` event,
-  not per-member), an explicit eventual-consistency tradeoff on an access-control
-  filter. That design decision is deferred to a follow-up rather than shipped
-  implicitly here.
+
+## Device/user Search scope + event-driven dynamic membership (follow-up, shipped)
+
+The `Search` RPC applied *no* device/user scope, and the device/user **list
+pages use `Search`** (not the scoped `ListDevices`/`ListUsers` RPCs) — so a scoped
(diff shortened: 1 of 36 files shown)


PaulDotterer:
@coderabbitai review

Issue #502, opened 2026-07-04
audit-log retention: tiered archival + snapshot-prune + crypto-shred erasure (SPEC-FIRST)


PaulDotterer:
2026.08 (compliance — the headline). **Large; reopens the rebuild core (#498) — full spec before code.**

The event store is append-only forever with no retention today. That's the GDPR storage-limitation gap and the SaaS-readiness gap. Design converged in discussion 2026-07-03.

## Rolling snapshot + prune
- [ ] Replay to a `sequence_num` checkpoint, persist projection state as the new "start," archive+prune events ≤ checkpoint. `RebuildAll` changes from truncate+replay-all → load-snapshot+replay-tail.
- [ ] Snapshot is a single global `sequence_num` cut (cross-projection consistency; `sequence_num` is monotonic so the cut is clean).
- [ ] Event-type-aware: high-frequency audit-only events (session-refresh) may hard-prune; domain events must be snapshotted/crypto-shredded, never lost, or replay breaks.

## Tamper-evidence preserved
- [ ] Pruning collides with the append-only trigger (migration 011 — a *security* control). The prune must emit its own `EventLogPruned{up_to_seq, archive_sha256, snapshot_id}` event so it lands IN the tamper-evident chain.
- [ ] The cold-storage archive export must be integrity-sealed (hash-chained / signed) — a plain editable SQL dump defeats the point.

## Tiered retention (industry norm — nobody deletes audit logs; compliance wants 1–7yr)
- [ ] Hot (queryable) → cold/archive (compressed SQL, cheap storage). Configurable retention per tier.

## Crypto-shred erasure (right-to-be-forgotten on an append-only store)
- [ ] Can't DELETE from append-only, can't lose replay — so destroy the actor→person linkage by crypto-erasing under the enc:v2/KEK model. The event row survives for replay; the PII becomes unreadable everywhere (live + cold archive) at once by destroying one key.
- [ ] GDPR note: audit logs have a legal-obligation exception to erasure (Art 17(3)(b)); actor is a pseudonymous ULID. Right applies to corporate users too (it's not absent — it's usually *overridden* by the controller's lawful basis). DPO sets periods — not legal advice.

## Doctor observability (report, don't enforce)
- [ ] `control doctor` REPORTS retention posture (configured window, oldest event, row count, "no policy configured" warning). It does NOT prune — destructive scheduled mutation belongs in a background worker (like stale-execution-expiry). Enforcement = worker; observation = doctor.

## Framing
[...]

PR #517, opened 2026-07-05
feat(retention): rolling snapshot + prune worker & snapshot-aware restore (spec 19 stage C2)

PaulDotterer:
Spec 19 (audit retention & PII erasure), **stage C2** — the rolling snapshot + prune worker and the snapshot-aware restore path, on top of the merged C1 foundation (#516: `ArchiveStore` + fs backend, `EventLogPruned`, the append-only trigger exemption, `PruneEventsUpTo`).

Closes part of `manchtools/power-manage-server#502`.

## What this delivers

- **Prune worker** (`internal/retention`): advisory-lock single-flighted, crash-resume idempotent, no-op when nothing is older than the window, archive-then-delete ordering (the sealed archive lands **before** any event is deleted).
- **The snapshot is the ciphertext events ≤ N** — not a projection dump. A prune seals `{events ≤ N}` (header + one `to_jsonb` event row per gzip'd line) through the streaming `ArchiveStore`, then deletes events ≤ N via the privileged prune path.
- **Snapshot-aware restore** (`store.RebuildAllFromArchive`): replays the archived events ≤ N followed by the live events > N through the **same appliers**, in one transaction. PII decrypts through the per-user DEK on the way in, so:
  - a crypto-shredded user reproduces as the redaction sentinel (erasure survives restore), and
  - **cold archives hold only ciphertext PII** — never a plaintext derived view.

## Why events, not a projection dump

An earlier revision serialized the decrypted projection tables into the archive. That put plaintext PII into cold storage and would let a restore reproduce PII even with the key table gone — defeating crypto-shred and contradicting the spec. The archive now holds the shreddable ciphertext source; restore re-derives state by replaying it.

## Acceptance criteria covered

- **AC 16/18/19/28** — checkpoint choice, archive-before-delete, `EventLogPruned` in the same tx.
- **AC 17** — full-fidelity: prune@N then restore (archived ≤ N + live > N) is **byte-identical** to a pre-prune `RebuildAll`, compared as a full-row dump of every `AllRebuildTargets` table (self-discovering).
- **AC 21 / 21a** — restore replays archived ≤ N + live > N; with the DEK table empty, **all** PII reproduces as the redaction sentinel (permanently unreadable, honoring erasure).
- **AC 22** — the archive is independently replayable offline (`ReadArtifact`), no live system.
- **AC 25/26/27** — no-op / crash-resume idempotent / single-flight.

## Notable fix

- **Migration 012** drops `security_alerts_projection_event_id_fkey` (a projection FK to `events` with no `ON DELETE`). Without it, `PruneEventsUpTo` fails with a foreign-key violation whenever an alert references a pruned event — a projection must not pin the prunable log. Regression test included.

## Not in this PR (stage D)

`control doctor` retention posture + dual key-invariant checks, the retention config RPCs, the CLI `control rebuild-projections` **auto-restore-from-archive** wiring, and the scheduler tick. A plain `RebuildAll` after a prune covers only events > N (spec-accepted/documented); the full-recovery entry point lands with the archive config in stage D.

## Verification

`go vet`, `gofmt`, `staticcheck`, and the `internal/store` + `internal/retention` + `internal/archtest` suites pass locally (real-Postgres testcontainers). `internal/retention` added to the `store-projectors` CI shard (the self-discovering coverage guard). Local CodeRabbit findings addressed (stream the artifact via `io.Pipe`; documented the exclusive replay boundary).

## Summary by CodeRabbit

* **New Features**
  * Added automated event-log retention pruning that archives prunable history to enable offline replay.
  * Added archive-based rebuild to restore projections from saved event history.

* **Bug Fixes**
  * Improved pruning correctness and safety around concurrent execution and already-pruned checkpoints.
  * Enhanced file writing to avoid losing close-related errors.

* **Tests**
  * Added comprehensive coverage for pruning/archiving, restore fidelity (including after pruning), idempotency, safety-margin behavior, and concurrent single-flight skipping.
  * Added checks for foreign-key constraints during prune.


PaulDotterer:
Re: the sequence-visibility race on `PruneCheckpointBefore` — verified against the schema and addressed in 8e517c2.

`events.sequence_num` is indeed a plain `nextval()` sequence (pre-commit, not commit-order-safe) and `occurred_at` is `DEFAULT now()` (transaction start). So the race is real in principle: a not-yet-committed append holding a lower `sequence_num` with an `occurred_at` before the cutoff could be deleted-but-never-archived.

In practice it's excluded because `occurred_at < now-window` means the event's transaction started a full retention window ago, and no append transaction stays open that long (appends are sub-second). But `NewWorker` only enforces `window > 0`, so a tiny/misconfigured window could reach it. Rather than depend on the stage-D config validation, I added a self-contained **1h safety floor** on the checkpoint: it only includes events whose `occurred_at` is older than `max(window, 1h)` ago, so every event ≤ N has certainly committed. Days/months windows dominate the floor (invisible in normal operation). Regression test `TestPrune_SafetyMarginProtectsRecentEvents` (1ns window → no-op because of the floor; RED without it).

PR #521, opened 2026-07-07
feat(inventory): server-scheduled inventory collection interval (spec 22)

### LINKED ISSUE #500: inventory: server-held collection interval (delivered on sync, like sync cadence)
Extracted from the old 2026.08 umbrella (now server#492, 2026.09).

Inventory cadence becomes a server-held setting delivered on the sync response — the exact `SetDeviceSyncInterval` / `sync_interval_minutes` pattern (per-device override, group minimum, default = current 24h). NOT an agent env var: cadence is policy; policy lives server-side, is audit-evented, and lets the server authoritatively flag "inventory overdue" even for offline devices (it knows the expected cadence). Principle: agent executes+reports, server evaluates+holds policy.

- [ ] Server setting + delivery on the sync response (mirror sync-interval).
- [ ] Agent honors the delivered interval for periodic baseline collection (on-connect refresh + manual `RefreshDeviceInventory` unchanged).
- [ ] Docs: freshness contract — server assertions from reported data are "as of last inventory"; grace periods absorb the cadence.

PaulDotterer:
Implements spec 22 (docs `06-specs/22-server-side-inventory-interval.md`) — closes #500 as re-pinned 2026-07-07: cadence is **server-held policy**, collection stays server-initiated over the existing WS4-signed `RequestInventory` path, **zero agent changes**.

## What

- **RPCs**: `SetDeviceInventoryInterval` / `SetDeviceGroupInventoryInterval` (clone of the sync-interval template): validate at boundary AND handler (`omitempty,gte=120,lte=10080`, 0 = inherit), `EnforceDeviceScopeOnBaseTier` / `EnforceDeviceGroupScope`, event-sourced and audit-visible.
- **Events/store**: `DeviceInventoryIntervalSet` / `DeviceGroupInventoryIntervalSet` + typed payloads + projector appliers (stale-replay guarded); migration 014 adds `inventory_interval_minutes` to both projections; full-fidelity rebuild fixture extended (byte-identical replay).
- **Resolution** (AC 3): device override > group MIN > 1440-minute default; 0 inherits at each level. The expression lives in `ListDeviceInventoryFreshnessBatch` and `ListStaleInventoryDevices`; tests pin both copies to identical semantics.
- **Scheduler** (`internal/inventorysched`, AC 5/6): fixed 15-min tick, advisory-lock single-flighted across replicas, at most one signed request per stale connected device per tick (`last_seen_at` within one tick), sub-tick Asynq deadline so requests to disconnected devices expire. **No per-tick batch cap (pinned)**. Panic-recovered wiring in `cmd/control`; gated by `CONTROL_INVENTORY_SCHEDULER_ENABLED` (default `true`; `false` logs one boot line and starts nothing — AC 10).
- **Freshness on the Device surface** (AC 7): `last_inventory_at` + `inventory_overdue` (resolved interval + `max(1h, 25%)` grace; never-collected devices age from `registered_at`) — computed from server-held policy, valid while offline; one batched freshness query per list page, no per-row N+1.
- `SignInventoryRequest` moved `api` → `taskqueue` so scheduler and manual `RefreshDeviceInventory` share one signing site.

## Deviations from the spec (spec amended in docs repo)

- Out-of-scope **mutations** return `PermissionDenied` (WS3 semantics of the cloned template; `NotFound` stays for absent targets and read visibility). The spec's rejection table said NotFound but also pinned "identical scope model to the sync-interval RPCs" — the template's security-reviewed behavior wins.
- Freshness is joined via one batched query per page (`ListDeviceInventoryFreshnessBatch`) rather than inlined into every `ListDevices*` row query — same no-N+1 guarantee, mirrors the existing assignment batch-map pattern in the same handler.
- No standalone `GetDeviceInventoryInterval` query: resolution is embedded in the two consumers above; a third copy would only serve tests and invite drift.

## Tests

Handlers (correct/absent/malformed/out-of-range/zero-inherit/unauthenticated/out-of-scope/not-found + event append, red-checked by guard neutralization), resolution + stale feed against real Postgres, scheduler (one-request-per-tick, single-flight via foreign advisory-lock session, fail-closed nil-signer and enqueue-failure), freshness boundary (29h vs 31h around interval+grace, offline device, never-collected), tick panic recovery, env gate, projector pure+integration, fidelity round-trip, `internal/inventorysched` added to the CI handlers-services shard (self-discovering guard caught it).

Gates: go vet, gofmt, staticcheck, full `go test ./... -count=1` green; local CodeRabbit review: 1 finding — aggregation pre-filter applied, suggested LIMIT skipped (pinned no-batch-cap).

## Summary by CodeRabbit

* **New Features**
  * Added configurable inventory collection intervals for individual devices and device groups.
  * Introduced a background scheduler that periodically sends inventory requests to stale devices.
  * Device views now show inventory freshness details, including last inventory time and whether inventory is overdue.

* **Bug Fixes**
  * Improved handling of inventory interval inheritance and stale-device detection.
  * Added safeguards so scheduler errors and panics do not stop future runs.

* **Documentation**
  * Updated control server docs with the new inventory scheduler setting.


PaulDotterer:
@coderabbitai review

PaulDotterer:
CodeRabbit re-review dispositions (all three rated Trivial):

1. **Duplicated resolution expression** — deliberate; `inventory_interval_query_test.go` pins both copies to identical semantics (CR concurs deferring is reasonable given sqlc's raw-SQL model).
2. **Advisory lock held across the sequential enqueue batch** — consequence of the pinned no-batch-cap decision (spec 22, out of scope). Revisit only if ingest pressure is observed.
3. **Freshness degrade-on-error indistinguishable from fresh** — acknowledged. The failure window is a warn-logged DB blip on the auxiliary freshness query; degrading to unset fields (UI shows "Never collected", no false overdue alarm) was the deliberate fail-soft choice. An explicit "freshness unavailable" state requires a new SDK proto field — deferred until a consumer needs to distinguish it.

Issue #475, opened 2026-06-28
Feature: per-device security-only-upgrade capability signal (gate `security_only` before dispatch)


PaulDotterer:
## Summary

Surface a per-device **"unattended-upgrades installed"** signal so the server/UI can validate (or warn) *before* dispatching a `security_only` package-upgrade action to devices that can't honor it.

Target: **2026.08** (carry-forward from the 2026.07 release tracker #320 — item (d)). This is a feature, not release polish, so it was split out.

## Background

The apt backend's security-only upgrade path requires the `unattended-upgrades` package. The agent now fails **closed** at execution time when it's missing — `ErrBackendUnavailable` with an actionable message, resolving the binary by absolute path first to survive a hardened-systemd `$PATH` (SDK `pkg/apt.go` `resolveUnattendedUpgrade`, carry-forward (c)).

That's correct fail-closed behavior, but it only surfaces **after dispatch**, per device, as a failed execution. An operator building a fleet-wide `security_only` upgrade has no way to know up front which targets lack `unattended-upgrades`.

Note this is apt-specific: dnf (`--security`) and zypper (`patch --category security`) support security-only natively; pacman/flatpak have no security channel and already fail closed with `ErrSecurityOnlyUnsupported`.

## Proposed

1. **Agent** reports a per-device capability/inventory signal for security-only upgrade support (at minimum: apt + `unattended-upgrades` present). Most naturally an inventory row or a package-manager capability flag alongside the existing inventory tables.
2. **Server** consumes it so a `security_only` upgrade action can be validated/annotated against its target set.
3. **Web** warns when a `security_only` action targets devices that can't honor it (pre-dispatch), instead of relying on post-hoc failed executions.

## Acceptance criteria (draft — refine in spec)

- [ ] A device exposes whether it can perform a security-only upgrade (and via which backend).
- [ ] Creating/editing a `security_only` upgrade action surfaces target devices that lack the capability.
- [ ] No regression to the agent's execution-time fail-closed guard (it stays as the backstop).

## Out of scope

- Auto-installing `unattended-upgrades` on the target (operator choice).
- Changing the agent execution-time guard (carry-forward (c) already landed it).

PaulDotterer:
Shipped as spec 23 (redesigned 2026-07-08 away from the pre-dispatch capability signal to a first-class execution status):

[...]

PR #522, opened 2026-07-09
feat: NOT_APPLICABLE execution status through the pipeline (spec 23)

PaulDotterer:
Implements the server side of spec 23: a `NOT_APPLICABLE` agent result flows inbox → `ExecutionNotApplicable` event → projector → `executions_projection` (status `not_applicable`, reason in the error column) → `GetExecution`/`ListExecutions`, mirroring the SKIPPED shape including the projection_version replay guard. No migration — the status column is unconstrained text and the execution stream is already a Go projector.

**Drive-by real bug fixed**: the new self-discovering enum-mapping parity test (`TestExecutionStatusMapping_RoundTripsEveryEnumValue`, iterates the proto enum registry) caught that `statusToString`/`stringToStatus` had **no case for SKIPPED** — a skipped execution came back as UNSPECIFIED from `GetExecution` and the SKIPPED list filter matched nothing. Fixed alongside; the parity test keeps every future enum value from repeating this.

Tests (all red-checked before implementation):
- `TestHandleExecutionResult_NotApplicable` — inbox → event → projection round-trip with reason (AC 4)
- `TestExecutionNotApplicableFromEvent_Pure` + `TerminalStates/NotApplicable` — projector decode + listener against real Postgres
- enum-mapping parity test (self-discovering, matches-zero guard)

Gate: go vet, gofmt, staticcheck, full suite (37 packages) green. Local CodeRabbit review: no findings.

Depends on manchtools/power-manage-sdk#304 (merged; go.mod bumped to it).

Refs manchtools/power-manage-server#475

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/manchtools/power-manage/server
 
-go 1.25.11
+go 1.25.12
 
 require (
 	connectrpc.com/connect v1.18.1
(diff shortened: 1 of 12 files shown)


PaulDotterer:
@coderabbitai review

PR #527, opened 2026-07-10
feat(api): ExportAuditEvents — chunked CSV/JSON audit-log export (spec 26)

### LINKED ISSUE #501: audit-log: richer web UI + filtering + export
2026.08 (auditability). The audit-log view is bare-bones and low-signal today.

- [ ] Surface actor / target / action / outcome / timestamp columns.
- [ ] Filtering — the `ListAuditEvents` request already supports actor_id / stream_type / event_type; expose them in the UI.
- [ ] The #496 events just added (osquery/log reads, LUKS-token grants, session lifecycle: Logout/RefreshToken) give the view far more to show — make sure they render meaningfully.
- [ ] Export (CSV/JSON) for DSAR / external review.
- [ ] Redaction stays enforced at the read boundary (already present) — verify the new event types are covered.

Pairs with the retention/erasure work (separate issue) and the existing Security-Audit + SIEM planning.

PaulDotterer:
Closes #501. SDK counterpart: manchtools/power-manage-sdk#309 (merged) + manchtools/power-manage-sdk#310 (TS wrapper). Web UI follows direct-to-main.

## What

Server side of spec 26: `ExportAuditEvents` — a **unary, chunked** export of the audit log as CSV or JSON for DSAR / external review. The client loops `next_page_token` until empty and concatenates chunks into one valid artifact.

**Why unary, not streaming:** both auth interceptors are deliberately fail-closed on streaming RPCs (`WrapStreamingHandler` → Unimplemented). A unary chunk loop reuses the entire existing auth/validate/authz/rate-limit chain unchanged; server memory stays bounded at one page (AC5).

- **Keyset pagination** on `sequence_num` (not OFFSET) — events appended mid-export can't shift rows into a later page and duplicate them. Page-size test seam (`SetExportPageSizeForTest`).
- **Redaction reused, not reimplemented** — export rows are built from `eventToProto`, the exact `ListAuditEvents` path; a test drives the REAL `CreateAction` emit and asserts the sentinel secret reaches neither format while `[REDACTED]` does.
- **Authz identical to the list** — `procedureAlternatives` gates the export by the `ListAuditEvents` permission (exclusively); the export can never widen access. `Export*` now also matches the expensive rate limiter.
- **Filters mirror the audit view** — exact actor, stream-type set, event-type substring (same ILIKE `!`-escaping), `occurred_at` range.

## Guard classifications (each conscious)

- `Export` added to `readOnlyRPCPrefixes` (event-append sweep) — the export appends no events.
- `ExportAuditEventsRequest.ActorId` shares the list filter's non-ULID exemption (system/device actors).
- `TestProcedureAlternatives_Exact` pin gains the export entry.

## Tests

9 handler tests against real Postgres: filters (stream set / actor / event-type / date range), chunk concatenation to one valid document with no duplicates, redaction in both formats, unsupported format, invalid page token, slice-cap regression (local CR finding), ILIKE metachar regression (red-checked by neutralizing the escaping in the query — fails for the right reason, restored via regen).

Local CodeRabbit review: 3 minor findings, all fixed (CSV row-count assertion, vacuous actor filter — the factories emit actor "test", so the test now drives a real handler as the probe actor — and the added LIKE-metachar case).

Verification gate: go vet ✓ staticcheck ✓ gofmt ✓ full `go test ./...` ✓ (workspace) + standalone `GOWORK=off go build` against merged sdk ✓.

## Summary by CodeRabbit

- **New Features**
  - Added audit-event exports in CSV and JSON formats.
  - Supports paginated downloads with actor, stream type, event type, and date-range filters.
  - Exported data follows the same permissions and redaction rules as audit-event listings.
  - JSON exports combine paginated results into a single valid document.

- **Documentation**
  - Documented export behavior and literal event-type filtering semantics.

- **Tests**
  - Added coverage for filtering, pagination, validation, formatting, and redaction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,6 +342,7 @@ OIDC identity provider management for SSO authentication.
 | Method | Description |
 |--------|-------------|
 | `ListAuditEvents` | Paginated event log. Filters: `actor_id`, `stream_type`, `event_type`. Returns raw event data from the event store. |
+| `ExportAuditEvents` | Chunked CSV/JSON export of the filtered log (spec 26). Unary keyset-paginated (client loops `next_page_token`, concatenates chunks into one valid artifact); same read-side redaction and same `ListAuditEvents` permission as the list. Filters: `actor_id`, `stream_types` set, `event_type` substring, `occurred_at` range. |
 
 ### Search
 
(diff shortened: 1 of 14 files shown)


PaulDotterer:
@coderabbitai review

PR #528, opened 2026-07-10
fix(search): O(1) last_seen_at touch on DeviceHeartbeat (#499)

### LINKED ISSUE #499: search: device online/offline stale up to 1h (heartbeat not indexed)
Extracted from the old 2026.08 umbrella (now server#492, 2026.09).

The devices page computes online/offline at query time against the **indexed** `last_seen_at` (`deviceStatusClause`, 5-minute window). But `AffectedSearchOps` has no `DeviceHeartbeat` case, so the indexed value only refreshes on other device events or the hourly reconcile — a healthy, heartbeating device shows **offline for up to an hour**. (`DeviceHeartbeat` events ARE appended by the inbox worker; Postgres is fresh — only the search copy is stale.)

- [ ] Fix: O(1) `HSET search:devices:<id> last_seen_at <ts>` touch on the heartbeat path (FT NUMERIC attribute updates automatically; optional per-device throttle). NOT a full row reload per heartbeat; NOT widening the 5-min window; NOT a faster full reconcile.
- [ ] Regression test: indexed `last_seen_at` advances after a `DeviceHeartbeat`; status filter returns the device online while it heartbeats.
- [ ] Keep the `UserLoggedIn`/`last_login_at` exclusion (sort-only, deliberate) — document the distinction in `AffectedSearchOps`.

PaulDotterer:
Closes #499.

## What

`AffectedSearchOps` had no `DeviceHeartbeat` case, so the indexed `last_seen_at` only refreshed on other device events or the hourly reconcile — a healthy, heartbeating device showed **offline in device search for up to an hour** while Postgres was fresh.

Exactly the fix shape the issue pins:

- **O(1) touch, not a reindex**: new `SearchOpTouchDeviceLastSeen` — the listener writes ONLY the `last_seen_at` hash field (FT NUMERIC attribute updates automatically) with the event's `occurred_at`, the same value the Postgres projection writes for that heartbeat. No row reload per heartbeat, no widened 5-min window, no faster reconcile.
- **Guarded Lua script**: a bare `HSET` on a missing key would create a partial hash that enters the FT index as a hostname-less ghost row; the script only touches EXISTING device rows — unindexed devices are picked up by the registration reindex / reconcile with a fresh value anyway.
- **`UserLoggedIn`/`last_login_at` exclusion kept** and the distinction documented in the classifier (status filter = correctness surface; last-login = sort-only).

## Tests

- Classifier: `DeviceHeartbeat` → touch op (table case).
- Listener e2e (real Postgres): heartbeat event fires the touch with `occurred_at`, and does **NOT** enqueue a reindex.
- Valkey integration (real valkey-bundle): a device with 2-hour-stale indexed `last_seen_at` fails the online-window query, re-enters it after the touch; touching an unindexed id creates **no key** and no ghost row.

Local CodeRabbit review: no findings. Verification gate: vet ✓ gofmt ✓ staticcheck ✓ full `go test ./...` ✓

## Summary by CodeRabbit

* **New Features**
  * Device heartbeat events now update a device’s “last seen” status in search results.
  * Devices can return to online search results promptly without a full data reload.
* **Bug Fixes**
  * Prevented missing or unindexed devices from appearing as incomplete (“ghost”) search results.
* **Tests**
  * Added coverage for heartbeat updates, online-status searches, and missing-device handling.

### DIFF
--- a/internal/api/search_listener.go
+++ b/internal/api/search_listener.go
@@ -24,11 +24,13 @@ import (
 //
 // Three methods, in order of how often they fire from this listener:
 //
+//   - TouchDeviceLastSeen — every device heartbeat (#499)
 //   - EnqueueReindex — every reindex op
 //   - EnqueueRemove  — every delete op
 //   - GetReverseMembers — only on remove for scopes with cascading
 //     parent rebuilds (action / action_set / definition)
 type SearchIndex interface {
+	TouchDeviceLastSeen(ctx context.Context, id string, lastSeen int64) error
 	EnqueueReindex(ctx context.Context, scope, id string, data *taskqueue.SearchEntityData) error
 	EnqueueRemove(ctx context.Context, scope, id string, cascadeIDs []string) error
 	GetReverseMembers(ctx context.Context, scope, id string) []string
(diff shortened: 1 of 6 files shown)


PaulDotterer:
@coderabbitai review

PR #538, opened 2026-07-12
fix: spec 29 first-batch runtime hardening — XFF right-to-left, control-URL validation, audit-redaction backstop

### LINKED ISSUE #537: spec 29 first-batch runtime hardening: XFF right-to-left, control-URL validation, audit-redaction fallback
Implements three of the five spec 29 first-batch findings (the pure-Go, fully test-verifiable ones):

- **AC 5-7** — Rate-limit client-IP resolution walks X-Forwarded-For right-to-left through trusted proxies (first untrusted address wins), instead of trusting the leftmost (spoofable) entry. Fixes both `clientIP` (Connect interceptor) and its sibling `ClientIPFromHTTP` (SCIM auth) via a shared walker.
- **AC 8-9** — Gateway `GATEWAY_CONTROL_URL` validated at startup (HTTPS + host, no user-info/query/fragment); rejected without echoing the raw value; success logs only the `scheme://host[:port]` origin.
- **AC 10-12** — Audit redactor applies a narrow recursive secret-key backstop for decoded objects with no redaction schema; byte-identical passthrough otherwise.

Deferred (own PRs / supervised): task-envelope HMAC binding (AC 1-4, coordinated queue-drain deploy), Traefik docker-socket-proxy (AC 13-14, live-ingress).

Spec: docs/content/06-specs/29-server-security-audit-hardening.md

PaulDotterer:
Closes #537. Implements three of the five spec 29 first-batch findings — the pure-Go, fully test-verifiable ones. Each is an independent defense-in-depth change committed separately.

## AC 5-7 — Right-to-left forwarded-IP resolution
`clientIP` (Connect interceptor) and its sibling `ClientIPFromHTTP` (SCIM auth) both selected the **leftmost** X-Forwarded-For entry, which a client behind a trusted proxy can spoof to poison per-IP rate-limit attribution. Extracted a single `resolveClientIP` walker used by both paths: walks XFF **right-to-left**, skipping trusted-proxy hops and returning the first untrusted address; a malformed hop or an all-trusted chain falls back to the direct peer; X-Real-IP is consulted only when XFF is absent. Both paths validate the resolved value with `net.ParseIP` so they cannot drift.

## AC 8-9 — Gateway control-URL validation
`GATEWAY_CONTROL_URL` was parsed only for `scheme==https` and then logged **raw** at startup (twice), so a mistyped/injected value carrying user-info/query/fragment credentials would land in logs. Added `config.ValidateControlURL`: HTTPS + host required; user-info/query/fragment rejected; the error names the bad component without echoing the raw value; success logs only `scheme://host[:port]`.

## AC 10-12 — Unknown-schema audit-redaction backstop
An audit event whose (stream,event) has no redaction schema was returned to the client verbatim, so a future/legacy raw-map append carrying a secret-looking key would leak it. When schema lookup finds nothing, recursively redact keys matching the narrow deny set derived from the codebase's **real** secret-field names (`password`/`passphrase`/`psk`, `private_key*`, `*_secret`/`*_hash`/`*_encrypted`/`*_enc`). The `*_encrypted`/`*_enc` suffixes specifically catch `client_secret_encrypted`/`secret_encrypted`/`private_key_enc`, which a `*_secret`/`*_hash`-only set misses. Re-marshals only when something changed, so non-secret objects stay byte-identical; schema-aware redaction is untouched.

## Deferred (not in this PR)
- **Task-envelope HMAC binding** (AC 1-4) — clean-break envelope requires a coordinated queue-drain deploy of control+gateway+indexer; needs a deploy window.
- **Traefik docker-socket-proxy** (AC 13-14) — changes the reference deployment's live ingress; can't be verified without a live Traefik.

Spec: `docs/content/06-specs/29-server-security-audit-hardening.md`. Full suite + `go vet ./...` green; TDD (red-before-green) for each change.

### DIFF
--- a/cmd/gateway/main.go
+++ b/cmd/gateway/main.go
@@ -147,13 +147,12 @@ func main() {
 		logger.Error("GATEWAY_VALKEY_ADDR is required")
 		os.Exit(1)
 	}
-	if cfg.ControlURL == "" {
-		logger.Error("GATEWAY_CONTROL_URL is required")
-		os.Exit(1)
-	}
-	controlURL, err := url.Parse(cfg.ControlURL)
-	if err != nil || controlURL.Scheme != "https" {
-		logger.Error("GATEWAY_CONTROL_URL must use https for the internal mTLS control connection", "control_url", cfg.ControlURL, "error", err)
+	// Validate the control URL and derive a non-secret origin for logs (spec 29
+	// AC8-9): reject non-HTTPS, hostless, or credential/query/fragment-bearing
+	// values without ever echoing the raw (possibly credential-laden) string.
+	controlOrigin, err := config.ValidateControlURL(cfg.ControlURL)
+	if err != nil {
+		logger.Error("invalid GATEWAY_CONTROL_URL", "error", err)
 		os.Exit(1)
 	}
 
@@ -219,7 +218,7 @@ func main() {
 	}
 
 	controlProxy := handler.NewControlProxy(controlHTTPClient, cfg.ControlURL, gatewayID)
-	logger.Info("control proxy initialized", "control_url", cfg.ControlURL)
+	logger.Info("control proxy initialized", "control_url", controlOrigin)
 
 	// Create connection manager
 	manager := connection.NewManager()
--- a/internal/api/audit_handler.go
+++ b/internal/api/audit_handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"log/slog"
 	"sort"
+	"strings"
 
 	"connectrpc.com/connect"
 	"google.golang.org/protobuf/types/known/timestamppb"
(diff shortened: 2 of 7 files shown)


PaulDotterer:
Thanks @greptileai — addressed both P2s:

**Connect-path `clientIP` not independently tested** — added `TestClientIP_ConnectPathResolution` (`internal/auth/interceptor_test.go`), which drives `clientIP` end-to-end through an httptest server + the real interceptor and asserts the echoed `clientIP(req)`: right-to-left resolution skips the trusted loopback tail and returns the real client, and a no-XFF trusted peer resolves to a parsable IP. This closes exactly the drift you flagged (wrong header/peer accessor).

**`_enc` suffix breadth (`audit_handler.go`)** — keeping the broad suffix by design. This backstop runs *only* for events with no redaction schema; `*_enc`/`*_encrypted` are part of the spec's deny set (spec 29 AC10), derived from the codebase's real ciphertext field names, and are deliberately self-discovering rather than a hardcoded `private_key_enc` match — the project prefers self-discovering guards over hardcoded lists that go stale, and over-redacting a non-secret field on an already-unclassified event is the fail-safe direction (it shows `[REDACTED]`, never leaks). Known-schema events are unaffected.

PR #542, opened 2026-07-12
fix: uniform PermissionDenied on out-of-scope execution/log reads (spec 29 S10)

### LINKED ISSUE #541: spec 29 S10 (exec/log): uniform PermissionDenied for load-then-scope handlers
Follow-up to #539/#540. `GetExecution`/`CancelExecution`/`GetDeviceLogResult` load the row (→ NotFound if missing) before the device-scope gate (→ PermissionDenied if out of scope), so a scope-restricted caller can distinguish an existing execution/log on a device outside their scope from a nonexistent id. These must load the row to learn its device, so 'scope-check first' is impossible; instead resolve the miss to match the out-of-scope path — a device_group-scoped caller gets PermissionDenied for both, a global caller keeps NotFound. Keeps the WS3 scope-denial convention (maintainer decision, 2026-07-12).

PaulDotterer:
Closes #541. Completes S10 (the user-group half shipped in #540).

`GetExecution`/`CancelExecution`/`GetDeviceLogResult` loaded the row (→ NotFound if missing) before the device-scope gate (→ PermissionDenied if out of scope), letting a scope-restricted caller confirm an execution/log-query id exists on a device outside their scope. Unlike the user-group handlers (whose scope keys on the id from the auth context), these must load the row to learn its device, so 'scope-check first' is impossible.

`deviceScopeMissError` resolves a miss to match the out-of-scope path: a device_group-scoped (restricted) caller gets a PermissionDenied **byte-identical** to `auth.EnforceDeviceScope`'s, so the two are indistinguishable; a global/unrestricted caller keeps the honest NotFound. Keeps the WS3 scope-denial convention (per maintainer decision) — no web error-code change.

New test `TestExecLogHandlers_ScopeExistenceOracleClosed` drives all three with a scoped caller against out-of-scope-existing and unknown ids (both PermissionDenied) and a global caller (NotFound). Also fixes GetDeviceLogResult's prior blanket 'any error → NotFound' to distinguish transient faults (now logged + Internal).

Spec: `docs/content/06-specs/29-...md` (S10).

### DIFF
--- a/internal/api/action_dispatch.go
+++ b/internal/api/action_dispatch.go
@@ -691,6 +691,12 @@ func (h *ActionHandler) GetExecution(ctx context.Context, req *connect.Request[p
 
 	exec, err := h.store.Repos().Execution.Get(ctx, req.Msg.Id)
 	if err != nil {
+		if store.IsNotFound(err) {
+			// Uniform with the out-of-scope path below (spec 29 S10): a
+			// scope-restricted caller must not tell a missing execution apart from
+			// one on a device outside their scope.
+			return nil, deviceScopeMissError(ctx, "GetExecution", ErrExecutionNotFound, "execution not found")
+		}
 		return nil, handleGetError(ctx, err, ErrExecutionNotFound, "execution not found")
 	}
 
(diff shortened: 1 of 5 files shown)


PaulDotterer:
Thanks @greptileai — both addressed:

- **`DeviceScopeListFilter` reused outside list context**: added `auth.IsDeviceScopeRestricted(ctx, permission)` and switched `deviceScopeMissError` to it, so the single-object access decision no longer borrows the list-filter's name.
- **`GetDeviceLogResult` out-of-scope existing not exercised**: the sub-test now seeds a real result on the out-of-scope device (via `Logs.CreateQueryResult`) and asserts PermissionDenied, so it confirms `EnforceDeviceScopeOnBaseTier` actually fires — the unknown-id assertion alone would have passed even with the scope check removed.

PR #544, opened 2026-07-12
fix: inbox worker requires agent-scheduled action to resolve to the device (spec 29 S5)

### LINKED ISSUE #543: spec 29 S5: inbox worker records execution/compliance for an unassigned agent-scheduled action
The agent-scheduled branch of `handleExecutionResult` looks the action up for existence only, never verifying it resolves (is assigned) to the reporting device. A compromised agent can forge ExecutionCreated/ExecutionCompleted/ComplianceResultUpdated (with self-reported `compliant`) for any action id that exists. Fix: gate the agent-scheduled path on `resolution.ResolveActionsForDevice` (the same call ProxySyncActions uses to sync the agent's action set), drop-with-WARN otherwise. A since-unassigned action is safe to drop — unassignment drives it to ABSENT and the agent rolls it back (maintainer confirmed 2026-07-12).

PaulDotterer:
Closes #543.

The agent-scheduled branch of `handleExecutionResult` (a result whose `action_id` is NOT a server-dispatched execution id) looked the action up for **existence only**, never checking it resolves to the reporting device. A compromised agent could therefore forge `ExecutionCreated`/`ExecutionCompleted`/`ComplianceResultUpdated` events — including a self-reported `compliant` — for any action id that exists in the org.

Fix: gate that path on `resolution.ResolveActionsForDevice` — the **same** call `ProxySyncActions` uses to sync the agent's action set — so a result is accepted only for an action currently assigned to the reporting device; otherwise it is dropped with a WARN.

**No legitimate results are lost.** The immediate drop is race-safe: projections are synchronous post-commit (`store.RegisterEventListener` fires before `AppendEvent` returns), so the read is read-your-writes consistent, and the agent only runs actions it was synced via the same resolution. The only thing that removes an action between run and result is an explicit unassignment — which drives the action to ABSENT and triggers the agent's rollback, so dropping its stale result is correct (confirmed with maintainer).

New test `TestHandleExecutionResult_AgentScheduled_RequiresAssignment`: an unassigned action's agent-scheduled result mints no execution; a REQUIRED-assigned action's result still records. Device-origin binding (existing) already blocks cross-device spoofing; this closes the same-device unassigned-action gap.

Spec: `docs/content/06-specs/29-...md` (S5).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of agent-scheduled execution results.
  * Results are now recorded only when the associated action is assigned to the reporting device.
  * Invalid or unassigned results are safely ignored, while temporary validation errors can be retried.

### DIFF
--- a/internal/control/inbox_worker.go
+++ b/internal/control/inbox_worker.go
@@ -23,6 +23,7 @@ import (
 	"github.com/manchtools/power-manage/server/internal/eventtypes"
 	"github.com/manchtools/power-manage/server/internal/eventtypes/payloads"
 	"github.com/manchtools/power-manage/server/internal/gateway/registry"
+	"github.com/manchtools/power-manage/server/internal/resolution"
 	"github.com/manchtools/power-manage/server/internal/store"
 	db "github.com/manchtools/power-manage/server/internal/store/generated"
 	"github.com/manchtools/power-manage/server/internal/taskqueue"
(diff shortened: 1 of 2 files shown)


PaulDotterer:
Good catch @greptileai — acknowledged as a bounded, accepted cost for now:

- `ResolveActionsForDevice` is the **same** resolution `ProxySyncActions` runs on every agent connect, over device-scoped indexed queries — so this doesn't add a new class of work, it repeats the per-connect cost per agent-scheduled result.
- Correctness (blocking forged execution/compliance for unassigned actions) is the point here, and the check must reflect exactly what the agent was synced.

A targeted single-action `IsActionResolvedForDevice(device, action)` query would cut it to one round-trip, but it would replicate the multi-layer (device/user/system) resolution UNION and risk drifting from `ResolveActionsForDevice` — so I'd only add it behind a test asserting the two agree. Filing that as a follow-up optimization rather than blocking the security fix.

PR #548, opened 2026-07-12
test: regression-test batch ID max=256 cap (spec 29 S7)

### LINKED ISSUE #546: spec 29 S7 (server): enforce + regression-test batch ID max=256
Companion to manchtools/power-manage-sdk#316. The repeated batch ID fields now carry max=256 in the SDK; every ControlService handler already runs api.Validate (proven by TestEveryControlRPCRunsValidateBeforeWork), which runs the go-playground validator over those tags — so the cap is enforced at the boundary with no server code change. Adds TestBatchIDFields_CappedAt256 as the regression test (256 passes, 257 rejected across required-batch + omitempty request variants). Depends on #316 being on sdk main for server CI.

PaulDotterer:
Closes #546. Companion to manchtools/power-manage-sdk#316 (merged).

The repeated batch ID request fields now carry `max=256` in the SDK. Every ControlService handler already runs `api.Validate` (proven by `TestEveryControlRPCRunsValidateBeforeWork`), which runs the go-playground validator over those `@gotags` — so the cap is enforced at the request boundary with **no server code change**.

`TestBatchIDFields_CappedAt256` pins it: a 256-id batch validates, a 257-id batch is rejected (CodeInvalidArgument), across the required-batch (`DispatchToMultipleRequest`) and omitempty (`CreateUserRequest`, `AssignRoleToUserGroupRequest`) variants. All eight capped fields share the identical tag + Validate path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Power Manage SDK integration to a newer version.

* **Tests**
  * Added validation coverage confirming batch requests accept up to 256 IDs and reject larger batches with an invalid-argument error.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/golang-jwt/jwt/v5 v5.3.1
 	github.com/hibiken/asynq v0.26.0
 	github.com/jackc/pgx/v5 v5.9.2
-	github.com/manchtools/power-manage-sdk v0.5.4-0.20260710151251-40990896bb10
+	github.com/manchtools/power-manage-sdk v0.5.4-0.20260712093419-4aeac8b4d5b8
 	github.com/oklog/ulid/v2 v2.1.0
 	github.com/pquerna/otp v1.5.0
 	github.com/pressly/goose/v3 v3.26.0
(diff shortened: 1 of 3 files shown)


PaulDotterer:
**Marked draft — blocked on the server's SDK dependency.** The S7 cap is enforced by the SDK (manchtools/power-manage-sdk#316, merged), but the server's `go.mod` pins the SDK at a 2026-07-10 pseudo-version (5 commits behind main, incl. unrelated #254/#310/#312/#314). This test therefore passes locally against the in-tree SDK but fails CI, which resolves the older pinned SDK where the cap doesn't exist yet. It'll go green once the server's SDK dependency is bumped to include #316 — which happens naturally with your next SDK release/tag. Not merging until then.

Also addressed @greptileai — the test now covers **all eight** capped fields (was 3), so a cap dropped from any one is caught; the 'only test 3' comment is corrected.

PR #551, opened 2026-07-12
fix: bind queued-task HMAC to version/direction/queue/type (spec 29 AC1-4)

### LINKED ISSUE #550: spec 29: bind queued-task HMAC to version/direction/queue/type (not just payload)
The Asynq task envelope signed ONLY the payload bytes, so a queue writer or holder of a valid signed task could replay those bytes under a different queue, task type, or direction — the consumer verified the payload then dispatched using the independently-supplied queue/type. Bind the HMAC to (version, direction, exact queue, task type, payload) via a clean-break v1 envelope: `[version][32-byte HMAC][payload]`, preimage = domain-tag || u8(version) || lp(direction) || lp(queue) || lp(type) || lp(payload), direction derived from the queue class (never caller-supplied). No payload-only fallback. Producers (EnqueueTo{Device,Control,Search}) fail closed on an unknown queue; VerifyMiddleware verifies against the actual queue (asynq ctx) + type. **Deploy note: drain pending Asynq queues (or stop producers) and deploy control+gateway+indexer from the same release — mixed old/new signers are intentionally unsupported.**

PaulDotterer:
Closes #550. The last spec-29 first-batch finding.

## ⚠️ Deploy coordination required (clean break)
This is an intentionally **incompatible** envelope format. Before deploying:
1. **Drain pending Asynq queues** (or stop producers) — in-flight old-format tasks will be rejected.
2. Deploy **control + gateway + indexer from the same release** — mixed old/new signers are unsupported.

## What
The Asynq task envelope signed **only the payload bytes**, so a queue writer (compromised Valkey) or a holder of a valid signed task could **replay** it under a different queue, task type, or direction — the consumer verified the payload and then dispatched using the independently-supplied Asynq queue/type.

New clean-break v1 envelope: `[1-byte version][32-byte HMAC][payload]`, HMAC preimage =
`domainTag || u8(version) || lp(direction) || lp(queue) || lp(type) || lp(payload)` (`lp` = u32be-length-prefixed). **Direction is derived from the queue class** (device:* → control→device, control:inbox → device→control, terminal-audit, search) — never caller-supplied. There is **no payload-only fallback**: a legacy envelope fails the version-byte check.

- Producers (`EnqueueTo{Device,Control,Search}`) bind the exact queue + type and **fail closed** on an unknown queue; the signed queue is applied last so a caller option can't reroute it.
- `VerifyMiddleware` verifies against the task's **actual** queue (from the Asynq context) + type, dead-lettering (SkipRetry) on any mismatch.

## Tests
Full rejection matrix (tamper signature/payload/version; replay across queue/direction/type; unknown queue; truncated; wrong key; **legacy payload-only envelope rejected**), a **real Asynq-server** test proving a queue/type replay never reaches the handler, and client/miniredis tests asserting the stored task verifies only under its actual queue+type. `taskqueue.WithQueue` is a queue seam for direct `mux.ProcessTask` tests (production runs through the real server, which supplies the queue). All of taskqueue/control/gateway/search green; `go vet ./...` clean.

### DIFF
--- a/internal/control/inbox_worker_test.go
+++ b/internal/control/inbox_worker_test.go
@@ -847,10 +847,17 @@ func TestInbox_RejectsCrossGatewayDeviceOrigin(t *testing.T) {
 			GatewayID:         claimedGateway,
 		})
 		require.NoError(t, err)
-		// Wrap exactly as the producer does so the mux's VerifyMiddleware
-		// admits the payload — the binding is the ONLY thing under test.
-		return asynq.NewTask(taskqueue.TypeExecutionResult, signer.Wrap(raw))
+		// Wrap exactly as the producer does (control:inbox queue, ExecutionResult
+		// type) so the mux's VerifyMiddleware admits the payload — the device→
+		// gateway binding is the ONLY thing under test here.
+		wrapped, err := signer.Wrap(taskqueue.ControlInboxQueue, taskqueue.TypeExecutionResult, raw)
+		require.NoError(t, err)
+		return asynq.NewTask(taskqueue.TypeExecutionResult, wrapped)
 	}
+
+	// inboxCtx carries the queue the VerifyMiddleware binds against (the real
+	// Asynq server would supply it).
+	inboxCtx := taskqueue.WithQueue(context.Background(), taskqueue.ControlInboxQueue)
 
 	seedVictimExecution := func(t *testing.T, st *store.Store, deviceID string) string {
 		t.Helper()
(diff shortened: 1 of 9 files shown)


PaulDotterer:
Thanks @greptileai — acknowledged. The 1.5s window is a *confirm-negative* wait, and it can't produce a spurious failure: a replayed task is rejected by VerifyMiddleware (SkipRetry → archived) **before** the handler, so it never reaches the `delivered` channel regardless of runner speed. The window only fails if a replay is *actually* delivered (a real binding bug), and the valid delivery has a generous 10s timeout. If it ever does flake on a slow runner, I'll switch the negative check to an Inspector assertion that both replays landed in the archived/dead set rather than a wall-clock wait.

PR #553, opened 2026-07-12
fix: refuse SCIM cross-provider email link of an already-owned user (spec 29 S14)

### LINKED ISSUE #552: spec 29 S14: refuse SCIM cross-provider email link of an already-owned user
The account-takeover guard at users_create.go only refuses linking when the matched user `HasPassword` (a local account). A **passwordless** user already linked to identity provider A (a higher-trust SSO IdP) passes the guard, so provider B — holding only a lower-trust SCIM token — can `POST /Users` by email and claim (manage/deprovision) A's user. Extend the guard: if the matched user already has an identity link to a DIFFERENT provider, refuse unless `trust_email_assertions` is set; fail closed if links can't be read. Compensating controls (per-provider tokens, verifyProviderOwnership) bound the rest. Maintainer chose to close (extend the guard) rather than accept, 2026-07-12.

PaulDotterer:
Closes #552. The `HasPassword` account-takeover guard only covers LOCAL accounts, so a **passwordless** user already owned by identity provider A (a higher-trust SSO IdP) could be silently claimed by provider B via a SCIM `POST /Users` by email — a lower-trust IdP's token seizing A's user. Extends the guard: if the matched user is already linked to a *different* provider, refuse unless `trust_email_assertions` is set; fail closed if links can't be read. Regression test across two providers (refused-without-trust / allowed-with-trust), red-checked. Per-provider tokens + `verifyProviderOwnership` bound the rest.

Note on the read-then-append (CR): the residual is a narrow concurrent-double-provision of a brand-new email, not the sequential takeover the guard closes (A's link is committed before a later B reads, projections being synchronous post-commit). A fully-atomic conditional invariant (multi-link is allowed under trust, so it's not a plain unique constraint) is disproportionate for this defense-in-depth guard; documented in-code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved account linking safeguards for passwordless users.
  - Prevented cross-provider account linking unless email assertions are explicitly trusted.
  - Added clearer conflict and error handling when an account is already linked elsewhere.

### DIFF
--- a/internal/scim/users_create.go
+++ b/internal/scim/users_create.go
@@ -116,6 +116,37 @@ func (h *Handler) createUser(w http.ResponseWriter, r *http.Request) {
 				writeError(w, http.StatusConflict, "email already belongs to a local account; cannot auto-link")
 				return
 			}
+
+			// S14 (spec 29): a passwordless user already OWNED by another identity
+			// provider must not be silently claimed by this one via email — a
+			// lower-trust IdP's SCIM token could otherwise seize a higher-trust
+			// IdP's passwordless SSO user (the HasPassword guard above only covers
+			// LOCAL accounts). Refuse unless the operator delegated identity to this
+			// provider (trust_email_assertions). Fail CLOSED if links can't be read.
+			//
+			// This read-then-append is not atomic, but the residual is narrow: it
+			// only affects two providers racing to create the FIRST link for the
+			// same brand-new email (→ a double-link), NOT the S14 threat of claiming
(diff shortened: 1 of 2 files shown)


PaulDotterer:
Closing as superseded by the consolidated implementation now on `main` (landed through #585). The current SCIM write path rejects cross-provider email auto-linking for an already-bound account unless `trust_email_assertions` is enabled, fails closed when the binding count cannot be read, and includes regression coverage for both rejection and trusted-assertion behavior. The pre-consolidation files changed by this PR no longer exist.

Issue #556, opened 2026-07-12
Gateway self-enrollment: control-side issuance, renewal, revocation, CRL (spec 31 PR2)


PaulDotterer:
Control-side of spec 31 (gateway self-enrollment + control-plane HA). Depends on sdk #318/#320.

- CA: `IssueGatewayCertificateFromCSR` (gateway peer class, fixed 45-day validity) via a shared `issueFromCSR`.
- Event-sourced `gateway` stream: `GatewayEnrolled`/`GatewayCertRenewed`/`GatewayRevoked` + typed payloads + Go projector + `gateways_projection` migration + sqlc.
- Handlers: `GatewayAuthService.EnrollGateway` (public, token-gated, rate-limited, constant-time), `InternalService.RenewGatewayCertificate` (peer-cert CN), `ControlService.GetCertificateRevocationList` (agent-facing) + `RevokeGatewayCertificate` + `ListGateways`.
- `RevokeGatewayCertificate`/`ListGateways` permissions (org-tier, admin-inherited); `gateway_not_found` error code.
- Config `CONTROL_GATEWAY_ENROLL_TOKEN`; mounting + rate limiters.

Follow-ups: PR3 gateway boot rewrite + Part C binding upgrade; PR4 control HA.

PR #559, opened 2026-07-12
feat: gateway enrollment client + in-memory mTLS + ServerAuth EKU fix (spec 31 PR3 foundation)

PaulDotterer:
## Summary

The **client-side building blocks** for gateway self-enrollment (spec 31 PR3), plus a **correctness fix** uncovered while wiring them. This is the foundation; the `cmd/gateway` boot rewrite + Part C binding flip follow (see *Remaining* — they need multi-process e2e).

### Correctness fix (behavior change)
- **`ca`**: gateway certs now carry **both `ClientAuth` AND `ServerAuth`** EKU. A gateway cert is a *client* to control's internal listener **and** the TLS *server* cert on its agent-facing listener; an agent verifies the gateway's server cert with the `ServerAuth` EKU, which a client-auth-only cert (PR2's version) would fail. Agent certs stay client-auth only. Regression-guarded: EKU assertions + an enrolled cert now `Verify`s as a TLS server cert against the returned CA.

### Building blocks (not yet wired)
- **`mtls`**: `NewServerTLSConfigFromPEM` (in-memory server mTLS) + `CertRotator` (atomic cert swap behind `GetCertificate`/`GetClientCertificate`) for zero-drop renewal.
- **`gwenroll`**: the gateway's enroll/renew client — keypair → CSR → `EnrollGateway`, reads `gateway_id` from the issued cert CN, renews over `InternalService` reusing the key for PoP. **End-to-end test drives the real enroll handler in-process.**

## Validation
`go vet`, `staticcheck`, `gofmt` clean; `ca`, `mtls`, `gwenroll` tests green (incl. the in-process enroll + ServerAuth-as-server-cert verification).

## Remaining (needs multi-process e2e — see PR description)
- `cmd/gateway` boot rewrite: enroll-on-boot, **remove static `--tls-cert`/`--tls-key`**, build TLS from the enrolled identity via `CertRotator`, renewal goroutine at 80% lifetime.
- **Part C**: flip the device→gateway binding (`registry.CheckDeviceGatewayBinding` + the `InternalService` callers + `control:inbox` worker) to read `gateway_id` from the authenticated **peer-cert CN** instead of the request body.
- **Open question blocking safe completion**: how the *agent* verifies the gateway's server cert (DNS-hostname vs CA/peer-class). The enrolled cert has a spiffe class SAN + CN=gateway_id but **no DNS SAN** (the CA rejects caller SANs), so DNS-based verification would break — this must be confirmed against the agent's gateway-connection TLS config and verified end-to-end before the boot rewrite lands.

Refs manchtools/power-manage-server#556

## Summary by CodeRabbit

- **New Features**
  - Added gateway self-enrollment and certificate renewal capabilities.
  - Gateway certificates now include the configured hostname, supporting secure server authentication.
  - Added automatic in-memory certificate rotation for ongoing mTLS connections.
  - Enforced TLS 1.3 for newly supported secure connections.

- **Bug Fixes**
  - Gateway certificate renewals now preserve the gateway hostname and appropriate authentication capabilities.

- **Tests**
  - Expanded coverage for enrollment, renewal, certificate validation, hostname verification, and invalid-token handling.


PaulDotterer:
Addressed review:
- **gwenroll.Renew nil-key panic** + **drops not_after**: `Renew` now guards a nil key, returns the new `not_after` (the renewal scheduler needs it), and updates the `Identity` in place so `CertPEM`/`TLSCert` don't go stale.
- **No CertRotator/NewServerTLSConfigFromPEM tests**: added `internal/mtls/inmemory_test.go` (config build + bad-CA/mismatched-key rejection; rotator atomic swap on both callbacks).
- **Renewal cert without DNS SAN** (CodeRabbit): added a WARN — a current cert lacking a DNS SAN predates the fix and the renewed cert would be agent-unverifiable; the operator should re-enroll (a restart away, identity being ephemeral-per-boot).
- **CertRotator zero-value**: `NewCertRotator` always seeds a cert; the zero value isn't a supported construction path (documented).

PR #583, opened 2026-08-03
refactor(config): replace the configuration file with environment variables

PaulDotterer:
Implements the operator decision of 2026-08-03 amending
`DESIGN_2026_07_31/00_TARGET_DESIGN.md` §11: control is configured entirely
through documented `POWER_MANAGE_`-prefixed environment variables. There is no
configuration file.

## What this delivers

The JSON loader, the `-config` flag, and `POWER_MANAGE_CONFIG_FILE` are
deleted. Every option is one `POWER_MANAGE_` variable declared in a single
table in `cmd/control/config.go`; the recognized set is derived from that
table by reflection, and startup fails closed on any `POWER_MANAGE_` variable
it does not declare, naming the variable. List options are comma-separated and
reject empty entries. All validation semantics and defaults are unchanged, and
the exported `Config` shape is byte-identical, so no downstream consumer
changed.

Secrets keep their file-referenced shape: `POWER_MANAGE_ENCRYPTION_KEY` /
`_FILE` and `POWER_MANAGE_SEALING_KEY` / `_FILE` accept exactly one of value
or file (both or neither fail closed — previously the direct value silently
won), and the session-signing key stays file-only PEM. Errors name variables
and never echo secret values; a test asserts the exclusion.

`deploy/setup.sh` renders `config/control.env` (mode 0600, enforced by
`validate_permissions`) instead of `config/control.json`, and compose consumes
it via `env_file` with no arguments to control. `.env.example` shrinks to the
four deployment values; `install.sh` and `smoke-test.sh` no longer emit the
dead `LOG_LEVEL`/`LOG_FORMAT` pass-throughs. `setup.sh` re-renders
`control.env` on every run (including through `deploy.sh`); QUICKSTART states
that honestly instead of silently preserving edits.

Prose in `cmd/control/README.md`, `deploy/QUICKSTART.md`, and `SECURITY.md`
describes only the implemented model; docref anchors were regenerated through
the CLI and new claims anchor the loader symbols.

## Verification observed on this HEAD

- `gofmt` clean, `go vet` clean, full `go test -count=1 ./...` green.
- Runtime fail-closed probes against the built binary: unknown
  `POWER_MANAGE_TYPO` aborts startup naming it; `-config` is rejected as an
  unexpected argument; a malformed list aborts naming
  `POWER_MANAGE_TRUSTED_PROXIES`.
- End-to-end boot under pure env configuration: fresh SQLite schema, both
  listeners ready, comma-separated CORS value parsed into both origins,
  `/ready` and `/health` ok, `bootstrap-admin` issues its single-use URL.
- `deploy/setup_test.sh` — pass (5 PASS, exit 0), including the negative
  assertion that `control.json` is no longer produced and a
  set-and-uniqueness check over the rendered variables.
- `docref check` — 15 up-to-date, 0 stale, 0 broken.
- Red-checks by scoped mutation (guard skip, list-parser skip): the matching
  tests failed for the right reason and passed after restore.
- Local `coderabbit review --uncommitted` over all 12 changed files — no
  findings.

## Notable consequences

- Exporting `POWER_MANAGE_RUN_SCALE_TEST` (the scale-test toggle) in an
  environment that starts `control` now fails startup. That is the fail-closed
  rule working as specified; the variable belongs to `go test`, not to the
  server process.
- The docs repository has a companion commit renaming
  `backup_max_lag`/`webhook_url` mentions to their variable names.

## Deliberately not in this branch

- No GitHub issue precedes this branch: issue creation was blocked by the
  local tooling permission gate; the PR body carries the full context instead.
- Operator edits to the rendered `config/control.env` do not survive
  `setup.sh`/`deploy.sh` re-rendering. Making them durable needs its own
  agreed mechanism; silently keeping a stale file was rejected because
  domain-derived values would rot with no guard.

## Summary by CodeRabbit

* **New Features**
  * Configure the control service entirely through documented `POWER_MANAGE_` environment variables.
  * Support inline or protected-file delivery for supported secrets.
  * Add validated `serve`, `bootstrap-admin`, and `backup-status` commands.

* **Bug Fixes**
  * Reject unknown, malformed, incomplete, or insecure configuration values during startup.
  * Ensure error messages never expose secret contents.

* **Documentation**
  * Updated quickstart and security guidance for environment-based configuration, protected secret files, and the new backup path.
  * Deployment setup now generates a secured environment file instead of JSON configuration.


PaulDotterer:
@coderabbitai review

PR #584, opened 2026-08-03
fix: harden control against five pen-test findings

PaulDotterer:
Five fixes from an adversarial pen test of the post-rewrite control server,
each verified against `DESIGN_2026_07_31/00_TARGET_DESIGN.md`. Every fix ships
a regression test confirmed to fail on the pre-fix code (red-check by scoped
mutation). One commit per finding, plus a docref-anchor refresh.

## Fixes

**`fix(dispatch)` — future-scheduled deliveries fired early (HIGH).**
`agentsync.Sync` listed PENDING/PUSHED rows with no availability filter and
called `MarkPushed` directly, bypassing the availability/epoch guard
`dispatcher.Dispatch` enforces. A dispatch with a future `run_at` was delivered
on the device's next routine SyncRequest and executed up to its full schedule
offset early, on every connected device. The guard is extracted to
`delivery.Sendable(row, epoch, now)` and shared by both paths; Sync skips a row
that is not yet sendable. A query-level time filter was rejected — it would drop
a PUSHED row awaiting reconnect redelivery on a newer epoch.

**`fix(idp)` — OIDC JIT stored un-normalized email (MEDIUM).**
JIT inserted the raw IdP email while SCIM and all lookups normalize; with a
case-sensitive email column but a `COLLATE NOCASE` unique index, a mixed-case
JIT subject was unfindable by lookup yet blocked re-insertion — unmanageable by
SCIM. JIT now normalizes on insert and on the cross-provider link guard.

**`fix(store)` — fuzzy-search CPU amplification (MEDIUM).**
The matcher iterated raw (non-deduped) query tokens while a short token disables
the trigram prefilter, so a ~1KB repeated-token query burned seconds of
single-core CPU per authenticated request. Query tokens are deduplicated once
(result-preserving; each token must match, so repeats are redundant).

**`fix(devicegroup)` — device existence oracle (LOW).**
`ListDeviceGroupsForDevice` returned OK+empty for an existing out-of-scope
device but NotFound for a nonexistent one. It now folds an out-of-scope device
to the same `device_not_found`, reusing `device.GetDevice`'s resolver; a genuine
resolution failure stays internal.

**`fix(secrets)` — at-rest secret AAD omitted the row discriminator (LOW).**
LPS/LUKS at-rest AAD bound only (device, action, type), so sibling rows (LPS by
username, LUKS by device_path) shared identical AAD and were interchangeable by
a DB-level attacker. `SecretAADForRow` binds the discriminator on every seal and
open — the agent-facing LUKS fetch included, which the original finding missed.
Reveal queries now return the discriminator (sqlc regenerated). No migration:
the datastore is created fresh.

## Verification observed on this HEAD

- `scripts/verify.sh` (gofmt, build, vet, staticcheck, full `go test`, docref) —
  green.
- Red-check by scoped mutation, all five: disabling each fix fails exactly its
  regression test (`TestAgentSync_DoesNotPullFutureScheduledDeliveriesForward`,
  `TestLinker_JITStoresNormalizedEmail`,
  `TestMatchFuzzyDocument_DeduplicatesRepeatedQueryTokens`,
  `TestListDeviceGroupsForDevice_OutOfScopeDeviceIsNotAnExistenceOracle`,
  `TestSecretAADForRow_BindsRowDiscriminator` +
  `TestAgentSecrets_LpsBatchIsAtomicAndUsernameBound`); restored, all pass.
- A pre-existing reveal-round-trip test fixture that sealed with the old
  three-segment AAD was updated to the row-discriminated form.
- Local `coderabbit review` was rate-limited (seat quota); relying on the PR
  CodeRabbit review.

## Summary by CodeRabbit

- **Security**
  - Strengthened protection for stored secrets by binding encryption to individual records.
  - Improved access controls when viewing device groups.

- **Improvements**
  - Synchronization now respects delivery schedules and retries eligible items on newer connections.
  - Email-based account linking and provisioning now consistently handle capitalization and surrounding spaces.
  - Fuzzy search provides more consistent results when queries repeat terms.

- **Documentation**
  - Updated the SCIM/OIDC/JIT integration reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ search semantics. PostgreSQL is gone and must not return.
 - The named protobuf RPCs remain the product contract except for the 14
   Gateway-only removals listed in the target.
 - Agents initiate one direct outbound mTLS stream to control.
-<!-- docref: begin src=internal/scim/users_write.go#Handler.provisionSubject:3b57e30f,internal/idp/linker.go#Linker.createUser:0b0b13c3,internal/identity/users.go#Handlers.EraseJITUser:6cc8f91a -->
+<!-- docref: begin src=internal/scim/users_write.go#Handler.provisionSubject:3b57e30f,internal/idp/linker.go#Linker.createUser:7858b2ad,internal/identity/users.go#Handlers.EraseJITUser:6cc8f91a -->
 - Human identity uses SCIM lifecycle management or optional per-provider OIDC
   JIT for homelabs. There is no manual user creation; JIT-created subjects have
   an explicit, provenance-gated erasure RPC. Bootstrap-admin is a one-time host
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ search semantics. PostgreSQL is gone and must not return.
 - The named protobuf RPCs remain the product contract except for the 14
   Gateway-only removals listed in the target.
 - Agents initiate one direct outbound mTLS stream to control.
-<!-- docref: begin src=internal/scim/users_write.go#Handler.provisionSubject:3b57e30f,internal/idp/linker.go#Linker.createUser:7f026f45,internal/identity/users.go#Handlers.EraseJITUser:6cc8f91a -->
+<!-- docref: begin src=internal/scim/users_write.go#Handler.provisionSubject:3b57e30f,internal/idp/linker.go#Linker.createUser:0b0b13c3,internal/identity/users.go#Handlers.EraseJITUser:6cc8f91a -->
 - Human identity uses SCIM lifecycle management or optional per-provider OIDC
   JIT for homelabs. There is no manual user creation; JIT-created subjects have
   an explicit, provenance-gated erasure RPC. Bootstrap-admin is a one-time host
--- a/internal/agentsecrets/service.go
+++ b/internal/agentsecrets/service.go
@@ -119,7 +119,8 @@ func (s *Service) GetLuksKey(ctx context.Context, deviceID string, request *pmv1
 	if !strings.HasPrefix(key.Passphrase, "enc:v1:") {
 		return nil, errors.New("LUKS passphrase is not encrypted at rest")
 	}
-	plaintext, err := s.atRest.DecryptWithContext(key.Passphrase, pmcrypto.SecretAAD(deviceID, request.ActionId, "luks"))
+	plaintext, err := s.atRest.DecryptWithContext(key.Passphrase,
+		pmcrypto.SecretAADForRow(deviceID, request.ActionId, "luks", key.DevicePath))
 	if err != nil {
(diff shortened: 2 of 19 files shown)


PaulDotterer:
CodeRabbit review disposition:

**Addressing:**
- **crypto AAD (Major)** — correct: the username/device_path discriminator does not distinguish sibling rotation rows sharing one identity, so ciphertext can still be swapped between rotations. Rebinding the at-rest AAD to the immutable per-row ULID id.
- **Empty canonical email (Minor)** — correct: a whitespace-only claim passes the `!= ""` gate but normalizes to `""`. Normalizing once and refusing an empty canonical email, with a regression test.

**Skipping, with reasons:**
- **Bounded delivery list ordering (Minor)** — the 1024-row `created_at` window is a pre-existing property this change did not introduce; a future PENDING row can occupy a slot only when a device has >1024 unfinished deliveries, and the availability-filtered periodic sweep still delivers due rows. Reworking the shared query's ordering (also used by WakeDevice) is out of scope for this fix.
- **Test seeds rows via raw SQL (Major)** — raw `f.raw.Exec` fixture seeding is the established pattern across the store test suite (130+ call sites); these tests assert Sync/reveal behavior, not audit coverage, so routing fixture setup through WithAudit adds ceremony without testing anything new.
- **Reveal audit-transaction structure (Major)** — the read→decrypt→audit ordering of RevealLpsPassword/RevealLuksKey is pre-existing and untouched by this change (which only adjusted the AAD). It is a legitimate hardening but a separate reveal-path refactor, not in scope for the AAD fix; flagging for a dedicated change.

PaulDotterer:
Second review round addressed:

**Fixed:**
- **AAD bound to the immutable row id** (was username/device_path) — closes ciphertext relocation between sibling rotation rows; reveal reads the row id, reveal queries reverted to id + secret (sqlc regenerated). Verified by a red-check that fails when the discriminator is ignored.
- **Empty canonical email** — auto-create now gates on the normalized email and resolves an empty address to the same unauthenticated no-matching-account answer (an earlier sentinel that mapped to Internal broke `TestSSOCallback_IgnoresAnUnverifiedEmailClaim`, caught by the full gate and corrected).
- **Linux username from the normalized email** — `createUser` now derives the username from the same normalized address it stores, so a padded valid email no longer yields an inconsistent username; regression test added.

**Skipping, with reasons:**
- **Fixture writes via raw SQL** — raw `f.raw.Exec` seeding is the established pattern across the store test suite (130+ sites); these tests assert Sync/reveal behavior, not audit coverage.
- **Migration for secrets under the old AAD** — per the workspace clean-break rule, this is a disposable pre-alpha datastore created fresh with no stored data; there is no prior ciphertext to migrate, so no compatibility path is added.

PaulDotterer:
@coderabbitai review

